PR #1, opened 2024-01-16
Spotless VS Code extension + gradle

tails618:
Set up Spotless VSC extension and Gradle setup


tails618:
ope it doesn't build :/

balghane (approved):

PR #3, opened 2024-01-20
PoseBuffer implementation copied from 2023 minibot

voidSevenSevenSix:
See title

### DIFF
--- a/src/main/java/frc/robot/utils/PoseBuffer.java
+++ b/src/main/java/frc/robot/utils/PoseBuffer.java
@@ -4,14 +4,13 @@ import java.util.Optional;
 
 import edu.wpi.first.math.Pair;
 import edu.wpi.first.math.geometry.Pose2d;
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/utils/PoseBuffer.java
```diff
@@ -0,0 +1,41 @@
+package frc.robot.utils;
+
+import java.util.Optional;
+
+import edu.wpi.first.math.Pair;
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.wpilibj.Timer;
+
+public class PoseBuffer {
+    RingBuffer<Pair<Double, Pose2d>> buffer = new RingBuffer<Pair<Double, Pose2d>>();
```
balghane: Can you put a comment on what's the double and what's the pose. And maybe a comment that all the timestamps should be sequential

balghane (approved):

PR #6, opened 2024-01-25
AvaEleanor/intake-class

avakimc:
- created intake.java, intakeCal.java, intakeConstants.java
- created baseline methods
- Eleanor wrote this description...

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -5,22 +5,18 @@
 package frc.robot;
 
 /**
- * The Constants class provides a convenient place for teams to hold robot-wide
- * numerical or boolean
- * constants. This class should not be used for any other purpose. All constants
- * should be declared
+ * The Constants class provides a convenient place for teams to hold robot-wide numerical or boolean
+ * constants. This class should not be used for any other purpose. All constants should be declared
  * globally (i.e. public static). Do not put anything functional in this class.
  *
- * <p>
- * It is advised to statically import this class (or one of its inner classes)
- * wherever the
+ * <p>It is advised to statically import this class (or one of its inner classes) wherever the
  * constants are needed, to reduce verbosity.
  */
 public final class Constants {
   public static class OperatorConstants {
     public static final int kDriverControllerPort = 0;
   }
-  
+
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
 
   public static final double PLACEHOLDER_DOUBLE = 0.0;
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -5,18 +5,22 @@
 package frc.robot;
 
 /**
- * The Constants class provides a convenient place for teams to hold robot-wide numerical or boolean
- * constants. This class should not be used for any other purpose. All constants should be declared
+ * The Constants class provides a convenient place for teams to hold robot-wide
+ * numerical or boolean
+ * constants. This class should not be used for any other purpose. All constants
+ * should be declared
  * globally (i.e. public static). Do not put anything functional in this class.
  *
- * <p>It is advised to statically import this class (or one of its inner classes) wherever the
+ * <p>
+ * It is advised to statically import this class (or one of its inner classes)
+ * wherever the
  * constants are needed, to reduce verbosity.
  */
 public final class Constants {
   public static class OperatorConstants {
     public static final int kDriverControllerPort = 0;
   }
-
+  
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
 
   public static final double PLACEHOLDER_DOUBLE = 0.0;
@@ -24,4 +28,7 @@ public final class Constants {
   public static final float PLACEHOLDER_FLOAT = 0;
 
   public static final int SPARK_INIT_RETRY_ATTEMPTS = 5;
+
+  /* time between compute cycles */
+  public static final double PERIOD_TIME_SECONDS = 0.02;
 }
--- a/src/main/java/frc/robot/RobotMap.java
+++ b/src/main/java/frc/robot/RobotMap.java
@@ -1,12 +1,12 @@
 package frc.robot;
 
 public class RobotMap {
-  public static final int PIGEON_CAN_ID = Constants.PLACEHOLDER_INT;
-  /** Driving TalonFX CAN IDs */
-  public static final int FRONT_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
-      REAR_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
-      FRONT_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
-      REAR_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT;
+    public static final int PIGEON_CAN_ID = Constants.PLACEHOLDER_INT;
+    /** Driving TalonFX CAN IDs */
+    public static final int FRONT_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
+            REAR_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
+            FRONT_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
+            REAR_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT;
 
   /** shooter SPARK MAX CAN IDS* */
   public static final int SHOOTER_MOTOR_A_CAN_ID = Constants.PLACEHOLDER_INT;
@@ -22,4 +22,9 @@ public class RobotMap {
 
   public static final int LEFT_ELEVATOR_CAN_ID = Constants.PLACEHOLDER_INT,
       RIGHT_ELEVATOR_CAN_ID = Constants.PLACEHOLDER_INT;
+
+    /** Intake motor and talon CAN IDs */
+    public static final int INTAKE_PIVOT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
+            INTAKING_FRONT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
+            INTAKING_BACK_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT;
 }
--- a/src/main/java/frc/robot/RobotMap.java
+++ b/src/main/java/frc/robot/RobotMap.java
@@ -1,12 +1,12 @@
 package frc.robot;
 
 public class RobotMap {
-    public static final int PIGEON_CAN_ID = Constants.PLACEHOLDER_INT;
-    /** Driving TalonFX CAN IDs */
-    public static final int FRONT_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
-            REAR_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
-            FRONT_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
-            REAR_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT;
+  public static final int PIGEON_CAN_ID = Constants.PLACEHOLDER_INT;
+  /** Driving TalonFX CAN IDs */
+  public static final int FRONT_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
+      REAR_LEFT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
+      FRONT_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT,
+      REAR_RIGHT_DRIVING_CAN_ID = Constants.PLACEHOLDER_INT;
 
   /** shooter SPARK MAX CAN IDS* */
   public static final int SHOOTER_MOTOR_A_CAN_ID = Constants.PLACEHOLDER_INT;
@@ -23,8 +23,8 @@ public class RobotMap {
   public static final int LEFT_ELEVATOR_CAN_ID = Constants.PLACEHOLDER_INT,
       RIGHT_ELEVATOR_CAN_ID = Constants.PLACEHOLDER_INT;
 
-    /** Intake motor and talon CAN IDs */
-    public static final int INTAKE_PIVOT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
-            INTAKING_FRONT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
-            INTAKING_BACK_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT;
+  /** Intake motor and talon CAN IDs */
+  public static final int INTAKE_PIVOT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
+      INTAKING_FRONT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
+      INTAKING_BACK_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT;
 }
--- a/src/main/java/frc/robot/subsystems/intake/Intake.java
+++ b/src/main/java/frc/robot/subsystems/intake/Intake.java
@@ -63,7 +63,7 @@ public class Intake extends SubsystemBase {
                         IntakeCal.PIVOT_MAX_VELOCITY_DEG_PER_SECOND,
                         IntakeCal.PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED));
 
-                        SparkMaxUtils.initWithRetry(this::setUpPivotSpark, IntakeCal.PIVOT_SPARK_INIT_RETRY_ATTEMPTS);
+                        SparkMaxUtils.initWithRetry(this::setUpPivotSpark, Constants.SPARK_INIT_RETRY_ATTEMPTS);
     }
 
     /** Does all the initialization for the pivot spark, return true on success */
--- a/src/main/java/frc/robot/subsystems/intake/Intake.java
+++ b/src/main/java/frc/robot/subsystems/intake/Intake.java
@@ -1,200 +1,211 @@
 package frc.robot.subsystems.intake;
 
-import java.util.Optional;
-import java.util.TreeMap;
-
 import com.ctre.phoenix6.configs.TalonFXConfiguration;
 import com.ctre.phoenix6.configs.TalonFXConfigurator;
 import com.ctre.phoenix6.controls.Follower;
 import com.ctre.phoenix6.hardware.TalonFX;
 import com.ctre.phoenix6.signals.NeutralModeValue;
 import com.revrobotics.AbsoluteEncoder;
+import com.revrobotics.CANSparkBase.IdleMode;
 import com.revrobotics.CANSparkLowLevel.MotorType;
 import com.revrobotics.CANSparkMax;
-import com.revrobotics.RelativeEncoder;
-import com.revrobotics.CANSparkBase.IdleMode;
 import com.revrobotics.SparkAbsoluteEncoder.Type;
-import edu.wpi.first.util.sendable.SendableBuilder;
 import edu.wpi.first.math.controller.ProfiledPIDController;
 import edu.wpi.first.math.trajectory.TrapezoidProfile;
 import edu.wpi.first.math.util.Units;
+import edu.wpi.first.util.sendable.SendableBuilder;
 import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
 import frc.robot.Constants;
 import frc.robot.RobotMap;
-import frc.robot.utils.SparkMaxUtils;
 import frc.robot.utils.SendableHelper;
+import frc.robot.utils.SparkMaxUtils;
+import java.util.Optional;
+import java.util.TreeMap;
 
 public class Intake extends SubsystemBase {
 
-    private final CANSparkMax pivotMotor = new CANSparkMax(RobotMap.INTAKE_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
-    private final AbsoluteEncoder pivotAbsoluteEncoder = pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
-    private final TalonFX intakeTalonFront = new TalonFX(RobotMap.INTAKING_FRONT_MOTOR_CAN_ID);;
-    private final TalonFX intakeTalonBack = new TalonFX(RobotMap.INTAKING_BACK_MOTOR_CAN_ID);
-    private final Follower intakeTalonBackFollower = new Follower(intakeTalonFront.getDeviceID(), false);
-    private ProfiledPIDController pivotController;
+  private final CANSparkMax pivotMotor =
+      new CANSparkMax(RobotMap.INTAKE_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+  private final AbsoluteEncoder pivotAbsoluteEncoder =
+      pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
+  private final TalonFX intakeTalonFront = new TalonFX(RobotMap.INTAKING_FRONT_MOTOR_CAN_ID);
+  ;
+  private final TalonFX intakeTalonBack = new TalonFX(RobotMap.INTAKING_BACK_MOTOR_CAN_ID);
+  private final Follower intakeTalonBackFollower =
+      new Follower(intakeTalonFront.getDeviceID(), false);
+  private ProfiledPIDController pivotController;
 
-    private Optional<Double> lastControlledTime;
-    private Optional<Double> prevVelocityDegPerSec;
+  private Optional<Double> lastControlledTime;
+  private Optional<Double> prevVelocityDegPerSec;
 
-    public enum IntakePosition {
-        DEPLOYED,
-        STOWED,
-        CLEAR_OF_CONVEYOR
-    }
+  public enum IntakePosition {
+    DEPLOYED,
+    STOWED,
+    CLEAR_OF_CONVEYOR
+  }
 
-    private TreeMap<IntakePosition, Double> intakePositionMap;
-    private IntakePosition desiredPosition = IntakePosition.STOWED;
+  private TreeMap<IntakePosition, Double> intakePositionMap;
+  private IntakePosition desiredPosition = IntakePosition.STOWED;
 
-    public Intake() {
-        initPivotMotor();
-        initIntakeTalons();
-        intakePositionMap = new TreeMap<IntakePosition, Double>();
-        intakePositionMap.put(IntakePosition.DEPLOYED, IntakeCal.INTAKE_DEPLOYED_POSITION_DEGREES);
-        intakePositionMap.put(IntakePosition.STOWED, IntakeCal.INTAKE_STOWED_POSITION_DEGREES);
-        intakePositionMap.put(IntakePosition.CLEAR_OF_CONVEYOR, IntakeCal.INTAKE_SAFE_POSITION_DEGREES);
-        lastControlledTime = Optional.of(Timer.getFPGATimestamp());
-    }
+  public Intake() {
+    initPivotMotor();
+    initIntakeTalons();
+    intakePositionMap = new TreeMap<IntakePosition, Double>();
+    intakePositionMap.put(IntakePosition.DEPLOYED, IntakeCal.INTAKE_DEPLOYED_POSITION_DEGREES);
+    intakePositionMap.put(IntakePosition.STOWED, IntakeCal.INTAKE_STOWED_POSITION_DEGREES);
+    intakePositionMap.put(IntakePosition.CLEAR_OF_CONVEYOR, IntakeCal.INTAKE_SAFE_POSITION_DEGREES);
+    lastControlledTime = Optional.of(Timer.getFPGATimestamp());
+  }
 
-    public void initPivotMotor() {
-        pivotController = new ProfiledPIDController(IntakeCal.INTAKE_PIVOT_P,
-                IntakeCal.INTAKE_PIVOT_I, IntakeCal.INTAKE_PIVOT_D, new TrapezoidProfile.Constraints(
-                        IntakeCal.PIVOT_MAX_VELOCITY_DEG_PER_SECOND,
-                        IntakeCal.PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED));
+  public void initPivotMotor() {
+    pivotController =
+        new ProfiledPIDController(
+            IntakeCal.INTAKE_PIVOT_P,
+            IntakeCal.INTAKE_PIVOT_I,
+            IntakeCal.INTAKE_PIVOT_D,
+            new TrapezoidProfile.Constraints(
+                IntakeCal.PIVOT_MAX_VELOCITY_DEG_PER_SECOND,
+                IntakeCal.PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED));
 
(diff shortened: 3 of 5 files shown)


src/main/java/frc/robot/RobotMap.java
```diff
@@ -13,4 +13,8 @@ public class RobotMap {
       REAR_LEFT_TURNING_CAN_ID = Constants.PLACEHOLDER_INT,
       FRONT_RIGHT_TURNING_CAN_ID = Constants.PLACEHOLDER_INT,
       REAR_RIGHT_TURNING_CAN_ID = Constants.PLACEHOLDER_INT;
+
+  /** Intake motor and talon CAN IDs */
+  public static final int INTAKE_PIVOT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
+      INTAKE_TALON_FRONT_CAN_ID = Constants.PLACEHOLDER_INT, INTAKE_TALON_BACK_CAN_ID = Constants.PLACEHOLDER_INT;
```
tails618: I would name these `INTAKING_FRONT_MOTOR_CAN_ID` or something just to make it clear what they actually do.

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.RobotMap;
+
+public class Intake extends SubsystemBase {
+
+    private final CANSparkMax pivotMotor = new CANSparkMax(RobotMap.INTAKE_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+    private final RelativeEncoder pivotEncoder = pivotMotor.getEncoder();
+    private final AbsoluteEncoder pivotAbsoluteEncoder = pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
+    private final TalonFX intakeTalonFront = new TalonFX(RobotMap.INTAKE_TALON_FRONT_CAN_ID);
+    private final Follower intakeTalonBack = new Follower(intakeTalonFront.getDeviceID(), false); // TODO check
```
tails618: I get that we're creating a follower and connecting it to the front motor, but where do we actually connect this Follower to the physical motor? I don't see us giving it the back CAN ID, or anything. (Admittedly, I just did a brief look at the docs and I didn't see anything either.)

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+        return Math.cos(
+                Units.degreesToRadians(
+                        pivotEncoder.getPosition() - IntakeConstants.INTAKE_POSITION_WHEN_HORIZONTAL_DEGREES));
+    }
+
+    public void moveToPos(IntakePosition pos) {
+        pivotController.setGoal(intakePositionMap.get(pos));
+        double intakeDemandVoltsA = pivotController.calculate(pivotEncoder.getPosition());
+        double intakeDemandVoltsB = IntakeCal.INTAKE_PIVOT_FEEDFORWARD
+                .calculate(pivotController.getSetpoint().velocity);
```
tails618: Pretty sure this should be position

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public boolean atDesiredIntakePosition(IntakePosition pos) {
+        double desiredPositionDegrees = intakePositionMap.get(pos);
+        double intakePositionDegrees = pivotEncoder.getPosition();
+        return Math.abs(intakePositionDegrees - desiredPositionDegrees) <= IntakeCal.INTAKE_MARGIN_DEGREES;
+    }
+
+    public boolean clearOfConveyorZone() {
+        return pivotEncoder.getPosition() > IntakeCal.CONVEYOR_ZONE_THRESHOLD_DEGREES;
```
tails618: Whether this is > or < depends on what 0deg is; have we defined that?

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,28 @@
...
+  /** PID values for the intake pivot motor */
+  public static final double INTAKE_PIVOT_P = Constants.PLACEHOLDER_DOUBLE,
+      INTAKE_PIVOT_I = Constants.PLACEHOLDER_DOUBLE, INTAKE_PIVOT_D = Constants.PLACEHOLDER_DOUBLE;
+
+  /** max velocity and acceleration (deg per second) for intake pivot motor */
+  public static final double PIVOT_MAX_VELOCITY_DEG_PER_SECOND = Constants.PLACEHOLDER_DOUBLE,
+      PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED = Constants.PLACEHOLDER_DOUBLE;
+
+  /** intake positions (degrees) */
+  public static final double INTAKE_DEPLOYED_POSITION_DEGREES = Constants.PLACEHOLDER_DOUBLE, INTAKE_STOWED_POSITION_DEGREES = Constants.PLACEHOLDER_DOUBLE, INTAKE_SAFE_POSITION_DEGREES = Constants.PLACEHOLDER_DOUBLE;
```
tails618: What's 0deg, and which way is positive? Add that to the comment, pls

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public static final double INTAKE_MARGIN_DEGREES = Constants.PLACEHOLDER_DOUBLE, CONVEYOR_ZONE_THRESHOLD_DEGREES = Constants.PLACEHOLDER_DOUBLE;
+
+  /** intake power [-1.0,1.0] */
+  public static final double INTAKING_POWER = Constants.PLACEHOLDER_DOUBLE, REVERSE_INTAKING_POWER = Constants.PLACEHOLDER_DOUBLE;
+
+  /** intake feedforward */
+  public static final SimpleMotorFeedforward INTAKE_PIVOT_FEEDFORWARD = new SimpleMotorFeedforward(Constants.PLACEHOLDER_DOUBLE,Constants.PLACEHOLDER_DOUBLE);
```
tails618: The values passed into `SimpleMotorFeedforward` should be their own variables.

vendordeps/Phoenix5.json
tails618, line 1: We shouldn't be using Phoenix5, I think. If we are using it, try to find a way to use Phoenix6 instead; if we aren't, remove the library.

tails618 (changes requested):

src/main/java/frc/robot/subsystems/intake/Intake.java
balghane: Yeah I think you need to make a TalonFX for the back and configure it and everything. I believe the Follower is something you send to the Talon with setControl from what I saw

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+    private final Follower intakeTalonBack = new Follower(intakeTalonFront.getDeviceID(), false); // TODO check
+    private ProfiledPIDController pivotController = new ProfiledPIDController(IntakeCal.INTAKE_PIVOT_P,
+            IntakeCal.INTAKE_PIVOT_I, IntakeCal.INTAKE_PIVOT_D, new TrapezoidProfile.Constraints(
+                    IntakeCal.PIVOT_MAX_VELOCITY_DEG_PER_SECOND,
+                    IntakeCal.PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED));
+
+    public enum IntakePosition {
+        DEPLOYED,
+        STOWED,
+        SAFE
```
balghane: Rename?

src/main/java/frc/robot/subsystems/intake/Intake.java
balghane: Velocity is right but let's use the acceleration too

https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/controller/SimpleMotorFeedforward.html

Maybe with
```
calculate​(double currentVelocity, double nextVelocity, double dtSeconds)
```

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+        double intakeDemandVoltsB = IntakeCal.INTAKE_PIVOT_FEEDFORWARD
+                .calculate(pivotController.getSetpoint().velocity);
+        double intakeDemandVoltsC = IntakeCal.ARBITRARY_INTAKE_PIVOT_FEEDFORWARD_VOLTS * getCosineArmAngle();
+        pivotMotor.setVoltage(intakeDemandVoltsA + intakeDemandVoltsB + intakeDemandVoltsC);
+
+        if (pos == IntakePosition.DEPLOYED) {
+            startRollers();
+        } else {
+            stopRollers();
+        }
```
balghane: Maybe put this in periodic, and we only start once it reaches deployed?

src/main/java/frc/robot/subsystems/intake/Intake.java
balghane: We should define and write that somewhere yeah

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,28 @@
...
+      INTAKE_PIVOT_I = Constants.PLACEHOLDER_DOUBLE, INTAKE_PIVOT_D = Constants.PLACEHOLDER_DOUBLE;
+
+  /** max velocity and acceleration (deg per second) for intake pivot motor */
+  public static final double PIVOT_MAX_VELOCITY_DEG_PER_SECOND = Constants.PLACEHOLDER_DOUBLE,
+      PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED = Constants.PLACEHOLDER_DOUBLE;
+
+  /** intake positions (degrees) */
+  public static final double INTAKE_DEPLOYED_POSITION_DEGREES = Constants.PLACEHOLDER_DOUBLE, INTAKE_STOWED_POSITION_DEGREES = Constants.PLACEHOLDER_DOUBLE, INTAKE_SAFE_POSITION_DEGREES = Constants.PLACEHOLDER_DOUBLE;
+
+  public static final double INTAKE_MARGIN_DEGREES = Constants.PLACEHOLDER_DOUBLE, CONVEYOR_ZONE_THRESHOLD_DEGREES = Constants.PLACEHOLDER_DOUBLE;
```
balghane: Do we need both `CONVEYOR_ZONE_THRESHOLD_DEGREES` and `INTAKE_SAFE_POSITION_DEGREES`? Actually yes probably we do

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,28 @@
...
+
+  /** intake power [-1.0,1.0] */
+  public static final double INTAKING_POWER = Constants.PLACEHOLDER_DOUBLE, REVERSE_INTAKING_POWER = Constants.PLACEHOLDER_DOUBLE;
```
balghane: We can probably just set these to 1.0 and -1.0 for now

vendordeps/Phoenix5.json
balghane: I think the code got updated to use Phoenix 6 and this file can just be removed

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.RobotMap;
+
+public class Intake extends SubsystemBase {
+
+    private final CANSparkMax pivotMotor = new CANSparkMax(RobotMap.INTAKE_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+    private final RelativeEncoder pivotEncoder = pivotMotor.getEncoder();
+    private final AbsoluteEncoder pivotAbsoluteEncoder = pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
+    private final TalonFX intakeTalonFront = new TalonFX(RobotMap.INTAKE_TALON_FRONT_CAN_ID);
```
balghane: We should configure everything with this. There's a lot of kinds of configuration https://api.ctr-electronics.com/phoenix6/release/java/com/ctre/phoenix6/configs/TalonFXConfiguration.html

balghane (changes requested):

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,92 @@
...
+import edu.wpi.first.math.trajectory.TrapezoidProfile;
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.RobotMap;
+
+public class Intake extends SubsystemBase {
+
+    private final CANSparkMax pivotMotor = new CANSparkMax(RobotMap.INTAKE_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+    private final RelativeEncoder pivotEncoder = pivotMotor.getEncoder();
+    private final AbsoluteEncoder pivotAbsoluteEncoder = pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
```
balghane: Shouldn't we be using this

src/main/java/frc/robot/subsystems/intake/Intake.java
tails618: Why velocity? Do we care about the velocity for the pivot?

avakimc: added acceleration. we need velocity because that is what a profile is i think

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: 90 degrees now represents when the intake is at the shooter direction, therefore this should still be a >

src/main/java/frc/robot/Constants.java
```diff
@@ -5,20 +5,26 @@
...
     public static final int kDriverControllerPort = 0;
   }
-
+  
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
 
   public static final double PLACEHOLDER_DOUBLE = 0.0;
   public static final int PLACEHOLDER_INT = 0;
+
+  public static final double PERIOD_TIME_SECONDS = 20.0;
```
balghane: Milliseconds, also needs a comment that this the time between compute cycles

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+    private Optional<Double> lastControlledTime;
+
+    public enum IntakePosition {
+        DEPLOYED,
+        STOWED,
+        CLEAR_OF_CONVEYOR
+    }
+
+    private TreeMap<IntakePosition, Double> intakePositionMap;
+    private IntakePosition latestPosition = IntakePosition.STOWED;
```
balghane: Are we using this

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public void initIntakeTalons() {
+        TalonFXConfigurator cfg = intakeTalonFront.getConfigurator();
+        TalonFXConfiguration toApply = new TalonFXConfiguration();
+        toApply.CurrentLimits.SupplyCurrentLimit = IntakeConstants.INTAKING_TALONS_CURRENT_LIMIT_AMPS;
+        toApply.CurrentLimits.SupplyCurrentLimitEnable = true;
+        toApply.MotorOutput.NeutralMode = NeutralModeValue.Coast;
+        cfg.apply(toApply);
+
+        intakeTalonBack.setControl(intakeTalonBackFollower);
```
balghane: Should do the same setup here as for the front one

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    public double getLastControlledTime() {
```
balghane: Comment on this function?

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+        if (lastControlledTime.isEmpty()) {
+            return Units.secondsToMilliseconds(Constants.PERIOD_TIME_SECONDS);
+        }
+        return lastControlledTime.get();
+    }
+
+    public void setLastControlledTime(double mostRecentTimestamp) {
+        lastControlledTime = Optional.of(mostRecentTimestamp);
```
balghane: Doesn't really need its own function probably

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+        pivotController.setGoal(intakePositionMap.get(pos));
+        double intakeDemandVoltsA = pivotController.calculate(getOffsetAbsPositionDeg());
+        double intakeDemandVoltsB = IntakeCal.INTAKE_PIVOT_FEEDFORWARD
+                .calculate(pivotController.getSetpoint().velocity);
+        double intakeDemandVoltsC = IntakeCal.ARBITRARY_INTAKE_PIVOT_FEEDFORWARD_VOLTS * getCosineArmAngle();
+
+        double currentTimestamp = Timer.getFPGATimestamp();
+        double timestampDifference = currentTimestamp - getLastControlledTime();
+        setLastControlledTime(currentTimestamp);
+        double intakeDemandVoltsD = IntakeCal.INTAKE_PIVOT_FEEDFORWARD.calculate(pivotController.getSetpoint().velocity,
```
balghane: Only need this one, does both V and A

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+                pivotController.getGoal().velocity, timestampDifference);
+
+        pivotMotor.setVoltage(intakeDemandVoltsA + intakeDemandVoltsB + intakeDemandVoltsC + intakeDemandVoltsD);
+    }
+
+    public boolean atDesiredIntakePosition(IntakePosition pos) {
```
balghane: this doesn't do desired, it checks for an input position. so the function name and the impl differ

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,42 @@
...
+            INTAKE_PIVOT_I = Constants.PLACEHOLDER_DOUBLE, INTAKE_PIVOT_D = Constants.PLACEHOLDER_DOUBLE;
+
+    /** max velocity and acceleration (deg per second) for intake pivot motor */
+    public static final double PIVOT_MAX_VELOCITY_DEG_PER_SECOND = Constants.PLACEHOLDER_DOUBLE,
+            PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED = Constants.PLACEHOLDER_DOUBLE;
+
+    /**
+     * intake positions (degrees) where 90 degrees represents the intake at the
+     * shooter
+     */
```
balghane: And zero is down

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final double INTAKE_MARGIN_DEGREES = Constants.PLACEHOLDER_DOUBLE,
+            CONVEYOR_ZONE_THRESHOLD_DEGREES = Constants.PLACEHOLDER_DOUBLE;
+
+    /** intake power [-1.0,1.0] */
+    public static final double INTAKING_POWER = 1.0,
+            REVERSE_INTAKING_POWER = -1.0;
+
+    /** intake feedforward */
+    public static final double FEEDFORWARD_STATIC_GAIN = Constants.PLACEHOLDER_DOUBLE,
+            FEEDFORWARD_VELOCITY_GAIN = Constants.PLACEHOLDER_DOUBLE;
```
balghane: These should have units?

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final SimpleMotorFeedforward INTAKE_PIVOT_FEEDFORWARD = new SimpleMotorFeedforward(
+            FEEDFORWARD_STATIC_GAIN, FEEDFORWARD_VELOCITY_GAIN);
+    public static final double ARBITRARY_INTAKE_PIVOT_FEEDFORWARD_VOLTS = Constants.PLACEHOLDER_DOUBLE;
+
+    /** */
```
balghane: What's the comment here

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,145 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Constants;
+import frc.robot.RobotMap;
+import frc.robot.subsystems.drive.DriveCal;
+import frc.robot.subsystems.drive.DriveConstants;
+
+public class Intake extends SubsystemBase {
+
+    private final CANSparkMax pivotMotor = new CANSparkMax(RobotMap.INTAKE_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
```
balghane: Where do we configure this

balghane (changes requested):

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: fine 😔

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: coolio

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
avakimc: whoopsies

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: ill add one for both

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
avakimc: gone.

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
tails618: You didn't actually fix this, you're still passing placeholders directly into the `SimpleMotorFeedforward`.

Edit: just saw you aren't done with revisions, sorry.

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,157 @@
...
+        cfgBack.apply(toApply);
+
+        intakeTalonBack.setControl(intakeTalonBackFollower);
+    }
+
+    /** @return the last controlled time by the timer, if the lastControlledTime is empty (first record), then return 20 milliseconds
+     */
+    public double getLastControlledTime() {
+        if (lastControlledTime.isEmpty()) {
+            return Constants.PERIOD_TIME_MILLISECONDS;
```
balghane: - I think if this is empty, then you want to use this period as the difference between timestamps, not as the start timestamp
- `IntakeCal.INTAKE_PIVOT_FEEDFORWARD.calculate` is going to expect seconds, so either change this variable to seconds (making it 0.02) or divide it by 1000 where you do use it

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,157 @@
...
+                        pivotEncoder.getPosition() - IntakeConstants.INTAKE_POSITION_WHEN_HORIZONTAL_DEGREES));
+    }
+
+    public double getOffsetAbsPositionDeg() {
+        return pivotAbsoluteEncoder.getPosition() + IntakeCal.INTAKE_ABSOLUTE_ENCODER_ZERO_OFFSET_DEG;
+      }
+
+    public void moveToPos(IntakePosition pos) {
+        pivotController.setGoal(intakePositionMap.get(pos));
+        double intakeDemandVoltsA = pivotController.calculate(getOffsetAbsPositionDeg());
```
balghane: Here we control based on the absolute encoder, but then in `getCosineArmAngle` we use the relative encoder. We should use the same encoder for both

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,157 @@
...
+    
+        double intakeDemandVoltsC = IntakeCal.ARBITRARY_INTAKE_PIVOT_FEEDFORWARD_VOLTS * getCosineArmAngle();
+
+        double currentTimestamp = Timer.getFPGATimestamp();
+        double timestampDifference = currentTimestamp - getLastControlledTime();
+        lastControlledTime = Optional.of(currentTimestamp);
+        double intakeDemandVoltsB = IntakeCal.INTAKE_PIVOT_FEEDFORWARD.calculate(pivotController.getSetpoint().velocity,
+                pivotController.getGoal().velocity, timestampDifference);
```
balghane: This isn't right, you need to use the previous controlled velocity for the first argument and the current velocity for the second velocity.

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,157 @@
...
+    public void reverseRollers() {
+        intakeTalonFront.set(IntakeCal.REVERSE_INTAKING_POWER);
+    }
+
+    public void periodic() {
+        if (atIntakePosition(IntakePosition.DEPLOYED)) {
+            startRollers();
+        } else {
+            stopRollers();
+        }
```
balghane: I think ultimately this won't be correct. We will have the intake out when we climb. So probably we should make commands explicitly call start and stop themselves.

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,40 @@
+package frc.robot.subsystems.intake;
+
+import edu.wpi.first.math.controller.SimpleMotorFeedforward;
+import frc.robot.Constants;
+
+public class IntakeCal {
+
+    /** PID values for the intake pivot motor */
+    public static final double INTAKE_PIVOT_P = Constants.PLACEHOLDER_DOUBLE,
+            INTAKE_PIVOT_I = Constants.PLACEHOLDER_DOUBLE, INTAKE_PIVOT_D = Constants.PLACEHOLDER_DOUBLE;
```
balghane: Input and output units?

src/main/java/frc/robot/subsystems/intake/IntakeCal.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final double INTAKE_MARGIN_DEGREES = Constants.PLACEHOLDER_DOUBLE,
+            CONVEYOR_ZONE_THRESHOLD_DEGREES = Constants.PLACEHOLDER_DOUBLE;
+
+    /** intake power [-1.0,1.0] */
+    public static final double INTAKING_POWER = 1.0,
+            REVERSE_INTAKING_POWER = -1.0;
+
+
+    public static final SimpleMotorFeedforward INTAKE_PIVOT_FEEDFORWARD = new SimpleMotorFeedforward(
+            Constants.PLACEHOLDER_DOUBLE, Constants.PLACEHOLDER_DOUBLE);
```
balghane: Comment with input and output units (degrees and volts)

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,157 @@
...
+
+import edu.wpi.first.math.controller.ProfiledPIDController;
+import edu.wpi.first.math.trajectory.TrapezoidProfile;
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Constants;
+import frc.robot.RobotMap;
+
+public class Intake extends SubsystemBase {
```
balghane, line 26: Can you send some random relevant stuff in Sendable?

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public void initPivotMotor() {
+        pivotController = new ProfiledPIDController(IntakeCal.INTAKE_PIVOT_P,
+            IntakeCal.INTAKE_PIVOT_I, IntakeCal.INTAKE_PIVOT_D, new TrapezoidProfile.Constraints(
+                    IntakeCal.PIVOT_MAX_VELOCITY_DEG_PER_SECOND,
+                    IntakeCal.PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED));
+        // inverted
+        // encoder gear ratio
+        // current limit
```
balghane: 🙂

balghane (changes requested):

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: - updated the method to return the timestamp difference, returning 20 milliseconds if lastControlledTime is empty
- changed variable to seconds

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: changed to absolute for both

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: so should I just delete this?

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public boolean atDesiredIntakePosition() {
+        double desiredPositionDegrees = intakePositionMap.get(desiredPosition);
+        double intakePositionDegrees = pivotEncoder.getPosition();
+        return Math.abs(intakePositionDegrees - desiredPositionDegrees) <= IntakeCal.INTAKE_MARGIN_DEGREES;
+    }
+
+    public boolean atIntakePosition(IntakePosition pos) {
+        double checkPositionDegrees = intakePositionMap.get(pos);
+        double intakePositionDegrees = pivotEncoder.getPosition();
```
balghane: We should use absolute encoder everywhere or relative everywhere

src/main/java/frc/robot/subsystems/intake/Intake.java
```diff
@@ -0,0 +1,205 @@
...
+            intakeDemandVoltsB = IntakeCal.INTAKE_PIVOT_FEEDFORWARD.calculate(prevVelocityMetersPerSec.get(),
+                    currentVelocity, getTimeDifference());
+        }
+
+        prevVelocityMetersPerSec = Optional.of(currentVelocity);
+
+        pivotMotor.setVoltage(intakeDemandVoltsA + intakeDemandVoltsB + intakeDemandVoltsC);
+    }
+
+    public boolean atDesiredIntakePosition() {
```
balghane: This has repeated code from the below function. Can you do this without repeating code

balghane (changes requested):

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: mb

src/main/java/frc/robot/subsystems/intake/Intake.java
avakimc: yupsies

balghane (approved):

tails618 (approved):

PR #9, opened 2024-01-28
Sofia sid/shooter

Panda8188:
s h o o t e r . j a v a
s h o o t e r c a l . j a v a
s h o o t e r c o n s t a n t s . ja v a

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -13,13 +13,13 @@ package frc.robot;
  * constants are needed, to reduce verbosity.
  */
 public final class Constants {
+  public static final double PLACEHOLDER_DOUBLE = 0.0;
+  public static final int PLACEHOLDER_INT = 0;
   public static class OperatorConstants {
     public static final int kDriverControllerPort = 0;
   }
-
+  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
 
-  public static final double PLACEHOLDER_DOUBLE = 0.0;
-  public static final int PLACEHOLDER_INT = 0;
-  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
+
 }
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -17,9 +17,10 @@ public final class Constants {
     public static final int kDriverControllerPort = 0;
   }
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
+  
   public static final double PLACEHOLDER_DOUBLE = 0.0;
   public static final int PLACEHOLDER_INT = 0;
   public static final float PLACEHOLDER_FLOAT = 0;
 
-  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
+  public static final int SPARK_INIT_RETRY_ATTEMPTS = 5;
 }
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -16,8 +16,9 @@ public final class Constants {
   public static class OperatorConstants {
     public static final int kDriverControllerPort = 0;
   }
+
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
-  
+
   public static final double PLACEHOLDER_DOUBLE = 0.0;
   public static final int PLACEHOLDER_INT = 0;
   public static final float PLACEHOLDER_FLOAT = 0;
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -13,13 +13,13 @@ package frc.robot;
  * constants are needed, to reduce verbosity.
  */
 public final class Constants {
-  public static final double PLACEHOLDER_DOUBLE = 0.0;
-  public static final int PLACEHOLDER_INT = 0;
   public static class OperatorConstants {
     public static final int kDriverControllerPort = 0;
   }
-  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
+  public static final double PLACEHOLDER_DOUBLE = 0.0;
+  public static final int PLACEHOLDER_INT = 0;
+  public static final float PLACEHOLDER_FLOAT = 0;
 
-
+  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
 }
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -21,4 +21,5 @@ public final class Constants {
 
   public static final double PLACEHOLDER_DOUBLE = 0.0;
   public static final int PLACEHOLDER_INT = 0;
+  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
 }
--- a/src/main/java/frc/robot/subsystems/shooter/Shooter.java
+++ b/src/main/java/frc/robot/subsystems/shooter/Shooter.java
@@ -12,17 +12,16 @@ import com.revrobotics.SparkPIDController.ArbFFUnits;
 import edu.wpi.first.math.controller.ProfiledPIDController;
 import edu.wpi.first.math.interpolation.InterpolatingDoubleTreeMap;
 import edu.wpi.first.math.trajectory.TrapezoidProfile;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
 import frc.robot.Constants;
 import frc.robot.RobotMap;
 
-public class Shooter {
+public class Shooter extends SubsystemBase{
 
   private final CANSparkMax motorA =
       new CANSparkMax(RobotMap.SHOOTER_MOTOR_A_CAN_ID, MotorType.kBrushless);
-  private final RelativeEncoder motorAEncoder = motorA.getEncoder();
   private final CANSparkMax motorB =
       new CANSparkMax(RobotMap.SHOOTER_MOTOR_B_CAN_ID, MotorType.kBrushless);
-  private final RelativeEncoder motorBEncoder = motorB.getEncoder();
   private final CANSparkMax pivotMotor =
       new CANSparkMax(RobotMap.SHOOTER_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
   private final AbsoluteEncoder pivotMotorAbsoluteEncoder =
@@ -53,9 +52,6 @@ public class Shooter {
     controllerB.setI(ShooterCal.MOTOR_B_kI);
     controllerB.setD(ShooterCal.MOTOR_B_kD);
     controllerB.setFF(ShooterCal.MOTOR_B_kFF);
-
-    
-
   }
 
   public void shoot(double speedA, double speedB) {
@@ -64,8 +60,8 @@ public class Shooter {
   }
 
   public void stop() {
-    motorA.stopMotor();
-    motorB.stopMotor();
+    controllerA.setReference(0, ControlType.kVelocity);
+    controllerB.setReference(0, ControlType.kVelocity);
   }
 
   /** Returns the cosine of the intake angle in degrees off of the horizontal. */
@@ -82,9 +78,23 @@ public class Shooter {
     double armDemandVoltsC = ShooterCal.ARBITRARY_PIVOT_FEED_FORWARD_VOLTS * getCosineArmAngle();
     pivotMotor.setVoltage(armDemandVoltsA + armDemandVoltsB + armDemandVoltsC);
   }
-  /** NOT DONE */
-  public void setLatchAngle() {
-    pivotMotor.set(ShooterConstants.LATCH_ANGLE_DEGREES);
+
+//**Shooter latches (post climbing) **/
+  public void latch(){
+    controlPosition(Constants.PLACEHOLDER_DOUBLE);
+  }
+
+  //** Puts shooter in position to latch (before climbing)**/
+  public void setLatchPosition() {
+    controlPosition(Constants.PLACEHOLDER_DOUBLE);
+  }
+//**Checks to see if shooter is in position to latch **/
+  public boolean checkLatch() {
+    if (pivotDesiredPosition == ShooterConstants.LATCH_ANGLE_DEGREES){
+      return true;
+    } else {
+      return false;
+    }
   }
 
 }
--- a/src/main/java/frc/robot/subsystems/shooter/Shooter.java
+++ b/src/main/java/frc/robot/subsystems/shooter/Shooter.java
@@ -77,18 +77,17 @@ public class Shooter extends SubsystemBase{
     double timestamp = Timer.getFPGATimestamp();
 
     double armDemandVoltsA = pivotController.calculate(getPivotPosition());
-    double armDemandVoltsB = pivotController.calculate(getPivotPosition());
     if (!prevTimestamp.isEmpty()) {
-      armDemandVoltsB += ShooterCal.PIVOT_MOTOR_FF.calculate(prevVelocityRPM, pivotController.getSetpoint().velocity, timestamp - prevTimestamp.get());
+      armDemandVoltsA += ShooterCal.PIVOT_MOTOR_FF.calculate(prevVelocityRPM, pivotController.getSetpoint().velocity, timestamp - prevTimestamp.get());
     } else {
-      armDemandVoltsB += ShooterCal.PIVOT_MOTOR_FF.calculate(pivotController.getSetpoint().velocity);
+      armDemandVoltsA += ShooterCal.PIVOT_MOTOR_FF.calculate(pivotController.getSetpoint().velocity);
     }
-    double armDemandVoltsC = ShooterCal.ARBITRARY_PIVOT_FEED_FORWARD_VOLTS * getCosineArmAngle();
+    double armDemandVoltsB = ShooterCal.ARBITRARY_PIVOT_FEED_FORWARD_VOLTS * getCosineArmAngle();
 
     prevTimestamp = Optional.of(timestamp);
     prevVelocityRPM = pivotController.getSetpoint().velocity;
     
-    pivotMotor.setVoltage(armDemandVoltsA + armDemandVoltsB + armDemandVoltsC);
+    pivotMotor.setVoltage(armDemandVoltsA + armDemandVoltsB);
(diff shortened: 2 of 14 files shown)


src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+  private final CANSparkMax motorB =
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_B_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax pivotMotor =
+      new CANSparkMax(RobotMap.SHOOTER_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+  private final AbsoluteEncoder pivotMotorAbsoluteEncoder =
+      pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
+  private double pivotDesiredPosition = ShooterCal.STARTING_POSITION_DEGREES;
+
+  /** Double distance, Double angle for pivotMotor */
+  private InterpolatingDoubleTreeMap pivotAngleMap;
```
balghane: Units in comment

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+  private final SparkPIDController controllerB;
+  private final ProfiledPIDController pivotController = new ProfiledPIDController(ShooterCal.PIVOT_MOTOR_kP, ShooterCal.PIVOT_MOTOR_kI, ShooterCal.PIVOT_MOTOR_kD, new TrapezoidProfile.Constraints(ShooterCal.PIVOT_MAX_VELOCITY_DEG_PER_SECOND, ShooterCal.PIVOT_MAX_ACCELERATION_DEG_PER_SECOND_SQUARED));
+
+  public Shooter() {
+    pivotAngleMap = new InterpolatingDoubleTreeMap();
+    pivotAngleMap.put(Constants.PLACEHOLDER_DOUBLE, Constants.PLACEHOLDER_DOUBLE);
+  pivotAngleMap.put(Constants.PLACEHOLDER_DOUBLE, Constants.PLACEHOLDER_DOUBLE);
+
+    controllerA = motorA.getPIDController();
+    controllerB = motorB.getPIDController();
```
balghane: Set these in declaration instead of constructor

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+
+    controllerA.setP(ShooterCal.MOTOR_A_kP);
+    controllerA.setI(ShooterCal.MOTOR_A_kI);
+    controllerA.setD(ShooterCal.MOTOR_A_kD);
+    controllerA.setFF(ShooterCal.MOTOR_A_kFF);
+
+    controllerB.setP(ShooterCal.MOTOR_B_kP);
+    controllerB.setI(ShooterCal.MOTOR_B_kI);
+    controllerB.setD(ShooterCal.MOTOR_B_kD);
+    controllerB.setFF(ShooterCal.MOTOR_B_kFF);
```
balghane: Move these into a init sparks function that checks revliberrors. Should also set coast mode and current limits.

balghane: And set velocity conversion factor between rpm and rps if you want to do rps

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+
+  /** Returns the cosine of the intake angle in degrees off of the horizontal. */
+  public double getCosineArmAngle() {
+    return Math.cos(
+        pivotMotorAbsoluteEncoder.getPosition() - ShooterConstants.POSITION_WHEN_HORIZONTAL_DEGREES);
+  }
+
+  private void controlPosition(double distance){
+    pivotController.setGoal(pivotAngleMap.get(distance));
+    pivotDesiredPosition = pivotAngleMap.get(distance);
```
balghane: Maybe only call `pivotAngleMap.get(distance)` once and give it a variable name "target pivot angle". Or swap the order of the lines and setGoal to the desired position

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+    double armDemandVoltsA = pivotController.calculate(pivotMotorAbsoluteEncoder.getPosition());
+    double armDemandVoltsB =
+        ShooterCal.PIVOT_MOTOR_FF.calculate(pivotController.getSetpoint().velocity);
```
balghane: Let's use acceleration as well, do what we did in the elevator PR with previous timestamp and velocity

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+    double armDemandVoltsC = ShooterCal.ARBITRARY_PIVOT_FEED_FORWARD_VOLTS * getCosineArmAngle();
+    pivotMotor.setVoltage(armDemandVoltsA + armDemandVoltsB + armDemandVoltsC);
+  }
+
+/** Set arm to latching position (up = 180deg). Modified controlPosition. **/
+  public void setLatchPosition(){
+    pivotController.setGoal(180);
+    pivotDesiredPosition = 180;
```
balghane: 180 isn't necessarily right. Move to a variable and make it a placeholder

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+    double armDemandVoltsA = pivotController.calculate(pivotMotorAbsoluteEncoder.getPosition());
+    double armDemandVoltsB =
+        ShooterCal.PIVOT_MOTOR_FF.calculate(pivotController.getSetpoint().velocity);
```
balghane: Same as before with acceleration.

src/main/java/frc/robot/subsystems/shooter/ShooterCal.java
```diff
@@ -0,0 +1,34 @@
+package frc.robot.subsystems.shooter;
+
+import edu.wpi.first.math.controller.SimpleMotorFeedforward;
+import frc.robot.Constants;
+
+public class ShooterCal {
+  public static final double MOTOR_A_kP = Constants.PLACEHOLDER_DOUBLE;
+  public static final double MOTOR_A_kI = Constants.PLACEHOLDER_DOUBLE;
+  public static final double MOTOR_A_kD = Constants.PLACEHOLDER_DOUBLE;
+  public static final double MOTOR_A_kFF = Constants.PLACEHOLDER_DOUBLE;
```
balghane: Can you combine these declarations and add a comment. Same for below sets
```
  /** Add a comment here about the units. Input RPM or RPS (whatever you're using for target units), output [-1,1] */
  public static final double MOTOR_A_kP = Constants.PLACEHOLDER_DOUBLE,
  MOTOR_A_kI = Constants.PLACEHOLDER_DOUBLE,
  MOTOR_A_kD = Constants.PLACEHOLDER_DOUBLE,
  MOTOR_A_kFF = Constants.PLACEHOLDER_DOUBLE;
```

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+public class Shooter extends SubsystemBase{
+
+  private final CANSparkMax motorA =
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_A_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax motorB =
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_B_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax pivotMotor =
+      new CANSparkMax(RobotMap.SHOOTER_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+  private final AbsoluteEncoder pivotMotorAbsoluteEncoder =
+      pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
```
balghane, line 29: `pivotMotorAbsoluteEncoder` will have an offset from zero, so anywhere we get the position we should offset. Maybe make a private method getPivotPosition that gets from the encoder and uses the offset

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,97 @@
...
+  public void setLatchPosition(){
+    pivotController.setGoal(180);
+    pivotDesiredPosition = 180;
+    double armDemandVoltsA = pivotController.calculate(pivotMotorAbsoluteEncoder.getPosition());
+    double armDemandVoltsB =
+        ShooterCal.PIVOT_MOTOR_FF.calculate(pivotController.getSetpoint().velocity);
+    double armDemandVoltsC = ShooterCal.ARBITRARY_PIVOT_FEED_FORWARD_VOLTS * getCosineArmAngle();
+    pivotMotor.setVoltage(armDemandVoltsA + armDemandVoltsB + armDemandVoltsC);
+  }
+  /** Check if robot is in latch position (180deg) */
```
balghane: This does not just check latch position so this comment isn't accurate

balghane (changes requested):

src/main/java/frc/robot/subsystems/shooter/Shooter.java
Panda8188: Switched to RPM instead of RPS (RPS was just a placeholder value because I didn't know what units to use

src/main/java/frc/robot/Constants.java
```diff
@@ -16,10 +16,10 @@ public final class Constants {
...
     public static final int kDriverControllerPort = 0;
   }
-
   public static final double NEO_FREE_SPEED_RPM = 5676.0;
-
   public static final double PLACEHOLDER_DOUBLE = 0.0;
   public static final int PLACEHOLDER_INT = 0;
   public static final float PLACEHOLDER_FLOAT = 0;
+
+  public static final int SPARK_INIT_RETRY_ATTEMPTS = PLACEHOLDER_INT;
```
balghane: We can just set this to 5

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
+package frc.robot.subsystems.shooter;
+
+import java.util.Optional;
+
+import javax.swing.text.html.Option;
```
balghane: Remove, probably an accidental auto-import

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private final CANSparkMax motorA =
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_A_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax motorB =
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_B_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax pivotMotor =
+      new CANSparkMax(RobotMap.SHOOTER_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+  private final AbsoluteEncoder pivotMotorAbsoluteEncoder =
+      pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
+  private double pivotDesiredPosition = ShooterCal.STARTING_POSITION_DEGREES;
```
balghane: Add units to variable name

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  public void shoot() {
+    controllerA.setReference(ShooterConstants.SHOOTER_MOTOR_SPEED_RPM, ControlType.kVelocity);
+    controllerB.setReference(ShooterConstants.SHOOTER_MOTOR_SPEED_RPM, ControlType.kVelocity);
+  }
+
+  public void stop() {
+    controllerA.setReference(0, ControlType.kVelocity);
+    controllerB.setReference(0, ControlType.kVelocity);
```
balghane: Can we just do `setVoltage` on the motor to 0. The current code will actively try to slow the shooter down, which is not needed and is a waste of battery. setVoltage to 0 will let it spin down at its own rate

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
...
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_A_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax motorB =
+      new CANSparkMax(RobotMap.SHOOTER_MOTOR_B_CAN_ID, MotorType.kBrushless);
+  private final CANSparkMax pivotMotor =
+      new CANSparkMax(RobotMap.SHOOTER_PIVOT_MOTOR_CAN_ID, MotorType.kBrushless);
+  private final AbsoluteEncoder pivotMotorAbsoluteEncoder =
+      pivotMotor.getAbsoluteEncoder(Type.kDutyCycle);
+  private double pivotDesiredPosition = ShooterCal.STARTING_POSITION_DEGREES;
+
+  private double prevVelocity = 0;
```
balghane: Units? Rpm right?

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
...
+  public void controlPosition(double distance){
+    pivotDesiredPosition = pivotAngleMap.get(distance);
+    pivotController.setGoal(pivotDesiredPosition);
+    double timestamp = Timer.getFPGATimestamp();
+
+    double armDemandVoltsA = pivotController.calculate(getPivotPosition());
+    double armDemandVoltsB = pivotController.calculate(getPivotPosition());
+    if (!prevTimestamp.isEmpty()) {
+      armDemandVoltsB += ShooterCal.PIVOT_MOTOR_FF.calculate(prevVelocity, pivotController.getSetpoint().velocity, timestamp - prevTimestamp.get());
+    }
```
balghane: Maybe `else` we do the `calculate` with just the `final double nextVelocityRpm = pivotController.getSetpoint().velocity`?

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
...
+    double armDemandVoltsC = ShooterCal.ARBITRARY_PIVOT_FEED_FORWARD_VOLTS * getCosineArmAngle();
+
+    prevTimestamp = Optional.of(timestamp);
+    prevVelocity = pivotController.getSetpoint().velocity;
+    
+    pivotMotor.setVoltage(armDemandVoltsA + armDemandVoltsB + armDemandVoltsC);
+  }
```
balghane: This seems like a lot of repeated code with `controlPosition`. Can you think of a way to reduce the repeated code?

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,150 @@
...
+  public boolean initSparks() {
+    int errors = 0;
+    errors += SparkMaxUtils.check(pivotMotor.restoreFactoryDefaults());
+    errors += SparkMaxUtils.check(motorA.restoreFactoryDefaults());
+    errors += SparkMaxUtils.check(motorB.restoreFactoryDefaults());
+
+    errors += SparkMaxUtils.check(pivotMotor.setIdleMode(IdleMode.kBrake));
+    errors += SparkMaxUtils.check(motorA.setIdleMode(IdleMode.kBrake));
+    errors += SparkMaxUtils.check(motorB.setIdleMode(IdleMode.kBrake));
```
balghane: Let's set these to coast so they don't slow down too quickly

src/main/java/frc/robot/subsystems/shooter/ShooterCal.java
```diff
@@ -0,0 +1,36 @@
+package frc.robot.subsystems.shooter;
+import edu.wpi.first.math.controller.SimpleMotorFeedforward;
+import frc.robot.Constants;
+
+public class ShooterCal {
+    /** Motor A & B & pivotMotor, Unit Inputs: RPM. output [-1,1] */
```
balghane: Can you add similar comments for the other sections? So that when people hover over `PIVOT_MOTOR_kI` for example, it'll show the comment

src/main/java/frc/robot/subsystems/shooter/ShooterCal.java
```diff
@@ -0,0 +1,36 @@
...
+    PIVOT_MOTOR_kI = Constants.PLACEHOLDER_DOUBLE,
+    PIVOT_MOTOR_kD = Constants.PLACEHOLDER_DOUBLE;
+    
+  public static final double PIVOT_MOTOR_KS = Constants.PLACEHOLDER_DOUBLE,
+    PIVOT_MOTOR_KV = Constants.PLACEHOLDER_DOUBLE,
+    PIVOT_MOTOR_KA = Constants.PLACEHOLDER_DOUBLE;
+  public static final SimpleMotorFeedforward PIVOT_MOTOR_FF = new SimpleMotorFeedforward(
+            PIVOT_MOTOR_KS, PIVOT_MOTOR_KV, PIVOT_MOTOR_KA);
+
+  public static final double ARBITRARY_FEED_FORWARD_VOLTS = Constants.PLACEHOLDER_DOUBLE;
```
balghane: Not used, remove?

src/test/java/PoseBufferTest.java
```diff
@@ -1,27 +1,27 @@
-import edu.wpi.first.math.geometry.Pose2d;
```
balghane: Can we un-remove the tests? It seems like sometimes it passes the unit tests on github and sometimes it does not. I saw the same issue on a PR I put up too

balghane (changes requested):

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,137 @@
...
+  /** Set arm to a specific angle using voltage
+   * @param angle 
+   * **/
+  public void controlPosition(double angle){
+    pivotDesiredPositionDegrees = angle;
+    pivotController.setGoal(angle);
+    double timestamp = Timer.getFPGATimestamp();
+
+    double armDemandVoltsA = pivotController.calculate(getPivotPosition());
+    double armDemandVoltsB = pivotController.calculate(getPivotPosition());
```
balghane: What going on here, why are we doing this twice

balghane (changes requested):

balghane (approved):

PR #12, opened 2024-02-03
CRT Util

voidSevenSevenSix:
Implemented CRT Util, added a unit test for it. I will add comments on the actual math later, I am leaving too soon to do that right now.

### DIFF
--- a/src/main/java/frc/robot/utils/CRTGearRatioUtil.java
+++ b/src/main/java/frc/robot/utils/CRTGearRatioUtil.java
@@ -3,12 +3,12 @@ package frc.robot.utils;
 import java.util.Optional;
 
 /* Helpful CRT functions - calculate the absolute position of an object attached to a gear
- * 
+ *
  * How to use
  * -Initialize the crtutil using the init function
- * -Then, to get the absolute location of the object attached to the gear, 
+ * -Then, to get the absolute location of the object attached to the gear,
  * call getAbsolutePositionOfMainObject on the util object with the values from the two gear
- * 
+ *
  */
 public class CRTGearRatioUtil {
   private final double mainGearRotationRatioTerm;
@@ -17,63 +17,69 @@ public class CRTGearRatioUtil {
   private final double secondaryGearRotationRatioTerm;
   private final int maxRotations;
   private final double mainGearCircumference;
-  private final double k; //kFactor
+  private final double k; // kFactor
   private final int verificationFractionDenominatorPerRotation = 6;
   private final double doubleDivisionErrorFactor = 0.05;
   private final double elevatorErrorFactor = 0.01;
 
   /**
    * Input units will affect output units of some functions
-   * 
-   * @param mainGearRotationRatioTerm      for every mainGearRotationRatioTerm
-   *                                       rotations of the main gear,
-   *                                       the secondary gear rotates
-   *                                       secondaryGearRotationRatioTerm times
-   * @param secondaryGearRotationRatioTerm for every mainGearRotationRatioTerm
-   *                                       rotations of the main gear,
-   *                                       the secondary gear rotates
-   *                                       secondaryGearRotationRatioTerm times
-   * @param maxRotations                   the desired maximum number of rotations that
-   *                                       the main gear will make that involve
-   *                                       unique positions of the object
-   *                                       attached to it
-   * @param mainGearCircumference          circumference of the main gear
-   * 
-   * @return Returns optional of the util if the gear ratio works with the number
-   *         of max rotations, otherwise returns an empty optional
+   *
+   * @param mainGearRotationRatioTerm for every mainGearRotationRatioTerm rotations of the main
+   *     gear, the secondary gear rotates secondaryGearRotationRatioTerm times
+   * @param secondaryGearRotationRatioTerm for every mainGearRotationRatioTerm rotations of the main
+   *     gear, the secondary gear rotates secondaryGearRotationRatioTerm times
+   * @param maxRotations the desired maximum number of rotations that the main gear will make that
+   *     involve unique positions of the object attached to it
+   * @param mainGearCircumference circumference of the main gear
+   * @return Returns optional of the util if the gear ratio works with the number of max rotations,
+   *     otherwise returns an empty optional
    */
-  public static Optional<CRTGearRatioUtil> init(int mainGearRotationRatioTerm, int secondaryGearRotationRatioTerm,
-      int maxRotations, double mainGearCircumference) {
-    CRTGearRatioUtil u = new CRTGearRatioUtil(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm, maxRotations,
-        mainGearCircumference);
-    if (u.validateGearRatioWithMaxRotations()) { //check if the gear ratio works with the required amount of rotations by the main gear
+  public static Optional<CRTGearRatioUtil> init(
+      int mainGearRotationRatioTerm,
+      int secondaryGearRotationRatioTerm,
+      int maxRotations,
+      double mainGearCircumference) {
+    CRTGearRatioUtil u =
+        new CRTGearRatioUtil(
+            mainGearRotationRatioTerm,
+            secondaryGearRotationRatioTerm,
+            maxRotations,
+            mainGearCircumference);
+    if (u.validateGearRatioWithMaxRotations()) { // check if the gear ratio works with the required
+      // amount of rotations by the main gear
       return Optional.of(u);
     }
     return Optional.empty();
   }
 
-  private CRTGearRatioUtil(int mainGearRotationRatioTerm, int secondaryGearRotationRatioTerm, int maxRotations,
+  private CRTGearRatioUtil(
+      int mainGearRotationRatioTerm,
+      int secondaryGearRotationRatioTerm,
+      int maxRotations,
       double mainGearCircumference) {
     this.mainGearRotationRatioTerm = (double) mainGearRotationRatioTerm;
     this.secondaryGearRotationRatioTerm = (double) secondaryGearRotationRatioTerm;
-    this.mainGearRotationRatioTermSimplified = (double) (mainGearRotationRatioTerm
-        / gcd(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm));
-    this.secondaryGearRotationRatioTermSimplfied = (double) (secondaryGearRotationRatioTerm
-        / gcd(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm));
+    this.mainGearRotationRatioTermSimplified =
+        (double)
+            (mainGearRotationRatioTerm
+                / gcd(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm));
+    this.secondaryGearRotationRatioTermSimplfied =
+        (double)
+            (secondaryGearRotationRatioTerm
+                / gcd(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm));
     this.maxRotations = maxRotations;
     this.mainGearCircumference = mainGearCircumference;
     this.k = kFactor();
   }
 
   /**
-   * 
    * @param a
    * @param b
    * @return Greatest common divisor of a and b
    */
   private static double gcd(double a, double b) {
-    if (b == 0.0)
-      return a;
+    if (b == 0.0) return a;
     return gcd(b, a % b);
   }
 
@@ -86,9 +92,8 @@ public class CRTGearRatioUtil {
   }
 
   /**
-   * 
    * @return Returns a multiplier used by CRT. You shouldn't need to use this function yourself, and
-   *         I don't fully understand it (Victor Chen did the math behind it).
+   *     I don't fully understand it (Victor Chen did the math behind it).
    */
   private int kFactor() {
     double sum = 0;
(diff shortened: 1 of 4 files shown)


src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+    return gcd(b, a % b);
+  }
+
+  public static double mod(double a, double b) {
+    double r = a % b;
+    if (r < 0) {
+      r += b;
+    }
+    return r;
+  }
```
balghane, line 92: Can you use this built-in func? https://stackoverflow.com/a/31140209

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+    this.secondaryGearRotationRatioTerm = (double) secondaryGearRotationRatioTerm;
+    this.mainGearRotationRatioTermSimplified = (double) (mainGearRotationRatioTerm
+        / gcd(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm));
+    this.secondaryGearRotationRatioTermSimplfied = (double) (secondaryGearRotationRatioTerm
+        / gcd(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm));
+    this.maxRotations = maxRotations;
+    this.mainGearCircumference = mainGearCircumference;
+  }
+
+  //"These two functions speak for themselves" - Hans Niemann commenting his code
```
balghane: What does gcd stand for. Also can probably be private? Or leave it public and put some unit tests for it.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+    if (r < 0) {
+      r += b;
+    }
+    return r;
+  }
+
+  /**
+   * 
+   * @return Returns a multiplier used by CRT. You shouldn't need to use this function yourself, and
+   *         I don't fully understand it (Victor Chen did the math behind it).
```
balghane: Make it private then

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+   */
+  public int kFactor() {
+    double sum = 0;
+    double a = secondaryGearRotationRatioTermSimplfied;
+    double b = mainGearRotationRatioTermSimplified;
+    for (int i = 1; i < b; i++) {
+      sum += ((a * (b - i) - 1.0) / b)
+          * (1 + Math.floor(((a * (b - i) - 1.0) / b)) - Math.ceil(((a * (b - i) - 1.0) / b)));
```
balghane: `(a * (b - i) - 1.0) / b)` this is used three times, maybe split it out into a variable. Also what is even happening here wow

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return (int) ((int) ((sum * b) + 1.0) / a); //basically, when you apply CRT, you can multiply by a certain factor (what this function returns, a constant for a given gear ratio) such that when using modding that factor times the output of crt and the simplified main gear rotation ratio term, it will correctly output a number between 0 and maxrotations. then, to account for wrap around and to make 0 full rotations work correctly, mod the main gear ratio term by the output + the simplified term
```
balghane: Is the behavior of `(int)` correct for you? Or do you want to round? https://stackoverflow.com/a/2143505

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * 
+   * @param mainGearRotationValue      Rotation value returned by the main gear
+   *                                   encoder
+   * @param secondaryGearRotationValue Rotation value returned by the secondary
+   *                                   gear encoder
+   * @return Returns the number of full rotations of the main gear
+   */
+  public int getMainGearFullRotations(double mainGearRotationValue, double secondaryGearRotationValue) {
+    double k = kFactor();
```
balghane: Maybe compute and save this in the constructor as a `private final` member variable, if it is regularly used but never changes after the constructor

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * 
+   * @return Returns true if the gear ratio works with the neccessary number of
+   *         max rotations, false otherwise
+   */
+  public boolean validateGearRatioWithMaxRotations() {
+    boolean works = true;
+    for (int i = 0; i < maxRotations * 6; i++) {
+      double elevatorPos = i * mainGearCircumference / 6; //set elevator positon (expected position that it will check with) through sixths of a rotation until max rotations is reached
```
balghane: Split out 6 into a named variable. Do you really need `mainGearCircumference` in here? Seems like you can just operate on main gear rotations.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+      double epoc = CRTGearRatioUtil.mod(elevatorPos / mainGearCircumference, 1) > 0.9 ? 0.0
+          : CRTGearRatioUtil.mod(elevatorPos / mainGearCircumference, 1) < 0.1 ? 0.0
+              : CRTGearRatioUtil.mod(elevatorPos / mainGearCircumference, 1); //calculate what the main gear encoder value would be at each position - the ternary operators account for annoying float division rounding problems on either end (0 and 1)
```
balghane: How did you pick 0.9 and 0.1. Can you split these out into a variable reasonably? Seems like they only work because you picked the number 6, and increments of 1/6 go 0 to 0.16 (larger than 0.1).

balghane: It seems like you actually just want to do "is approx 0" or "is approx 1" and this isn't a great way to do that. Also this would be clearer if you had ` CRTGearRatioUtil.mod(elevatorPos / mainGearCircumference, 1)` once instead of three times.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+      if (Math.abs(this.getAbsolutePositionOfMainObject(epoc, //the calculated main gear encoder value
+          CRTGearRatioUtil.mod(
+              elevatorPos / mainGearCircumference * secondaryGearRotationRatioTerm /
+                  mainGearRotationRatioTerm, //the calculated secondary gear encoder value
+              1))
```
balghane: Why does this not need the weird "accounting for nearly 0 or 1" that you did above?

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+          - elevatorPos) > 0.01) { //check if the difference in what the function returns is greater than a small error factor, if it is, set working to false - this is a result more annoying float division problems, but will also not work if the gear ratio doesn't work
```
balghane: Make `0.01` a constant, call it error margin or somthing

balghane: I'm confused about this validation approach because can't you just check if the desired max travel is greater than `wrapAroundAbsolutePosition`?

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+   *                                       the secondary gear rotates
+   *                                       secondaryGearRotationRatioTerm times
+   * @param secondaryGearRotationRatioTerm for every mainGearRotationRatioTerm
+   *                                       rotations of the main gear,
+   *                                       the secondary gear rotates
+   *                                       secondaryGearRotationRatioTerm times
+   * @param maxRotations                   the maximum number of rotations that
+   *                                       the main gear will make that involve
+   *                                       unique positions of the object
+   *                                       attached to it
```
balghane: Is this the desired max rotations, or the max rotations that the gear ratio allows? A little confused here. Double check how this is used because it seems like it's being used for both?

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return works;
+  }
+
+  /**
+   * 
+   * @param mainGearRotationValue
+   * @param secondaryGearRotationValue
+   * @return Applies CRT to find the absolute position of the object attached to
+   *         the main gear
```
balghane: Note units are based on input circumference

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+   *                                       secondaryGearRotationRatioTerm times
+   * @param secondaryGearRotationRatioTerm for every mainGearRotationRatioTerm
+   *                                       rotations of the main gear,
+   *                                       the secondary gear rotates
+   *                                       secondaryGearRotationRatioTerm times
+   * @param maxRotations                   the maximum number of rotations that
+   *                                       the main gear will make that involve
+   *                                       unique positions of the object
+   *                                       attached to it
+   * @param mainGearCircumference          circumference of the main gear
```
balghane: Note the input units will set the output units for some of the functions

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+        (mainGearRotationValue * secondaryGearRotationRatioTermSimplfied
+            - secondaryGearRotationValue * mainGearRotationRatioTermSimplified) //this returns a decimal where the multiplier is different for every gear ratio
+            , mainGearRotationRatioTermSimplified) //mod it with the main gear rotation ratio term so it is within the correct range
+            + mainGearRotationRatioTermSimplified)); //for more information, see the comment in the kFactor() function
+  }
+
+  /**
+   * 
+   * @return Returns true if the gear ratio works with the neccessary number of
+   *         max rotations, false otherwise
```
balghane: Here max rotations is used as the desired max

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
```diff
@@ -0,0 +1,154 @@
...
+  public double getAbsolutePositionOfMainObject(double mainGearRotationValue, double secondaryGearRotationValue) {
+    return mainGearRotationValue * mainGearCircumference
+        + getMainGearFullRotations(mainGearRotationValue, secondaryGearRotationValue) * mainGearCircumference; //multiplies the number of full rotations plus the fractional amount of rotation from the main gear encoder by the circumference to determine distance traveled by the object attached to the main gear
+  }
+
+  /**
+   * @return Returns the absolute position of the main object after one wrap around
+   */
+  public double wrapAroundAbsolutePosition(){
+    return mainGearCircumference*maxRotations;
```
balghane: Here max rotations is used as the actual max that the gear ratio setup allows

src/test/java/CRTGearRatioUtilTestingStuff.java
```diff
@@ -0,0 +1,43 @@
...
+  @Test
+  void test() {
+    assert checker(44, 42, 22);
+    assert checker(44, 26, 22);
+    assert !checker(3, 4, 22);
+    assert !checker(4,3,10);
+    assert checker(28,30,14);
+  }
+  /*Main gear ratio, secondary gear ratio, max rotations */
+  boolean checker(int mgr, int sgr, int mr){
```
balghane: Fix function comment format. These should be `@param` and there should be a main function comment that says what this is doing. Returns false if the arguments are an invalid CRT setup

src/test/java/CRTGearRatioUtilTestingStuff.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Optional;
+
+public class CRTGearRatioUtilTestingStuff {
+  @Test
+  void test() {
+    assert checker(44, 42, 22);
+    assert checker(44, 26, 22);
+    assert !checker(3, 4, 22);
+    assert !checker(4,3,10);
+    assert checker(28,30,14);
```
balghane: Can you leave a comment on some of these stating why we should believe that it should work or should not work, whatever we're asserting?

src/test/java/CRTGearRatioUtilTestingStuff.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+  /*Main gear ratio, secondary gear ratio, max rotations */
+  boolean checker(int mgr, int sgr, int mr){
+    int mainGearRotationRatioTerm = mgr;
+    int secondaryGearRotationRatioTerm = sgr;
+    int maxRotations = mr;
+    double mainGearCircumference = Math.PI;
+    Optional<CRTGearRatioUtil> crtutilOptional = CRTGearRatioUtil.init(mainGearRotationRatioTerm, secondaryGearRotationRatioTerm, maxRotations,
+        mainGearCircumference);
```
balghane: capital U in Util

src/test/java/CRTGearRatioUtilTestingStuff.java
```diff
@@ -0,0 +1,43 @@
...
+     : CRTGearRatioUtil.mod(elevatorPos / mainGearCircumference, 1);
+     if(Math.abs(crtutil.getAbsolutePositionOfMainObject(epoc,
+      CRTGearRatioUtil.mod(
+      elevatorPos / mainGearCircumference * secondaryGearRotationRatioTerm /
+      mainGearRotationRatioTerm,
+      1)) - elevatorPos) > 0.01){
+        works = false;
+      }
+     }
+     return works;
```
balghane: Is this whole thing basically copied from the util? Can we find a way to reduce code redundancy

balghane (changes requested):

src/test/java/CRTGearRatioUtilTestingStuff.java
```diff
@@ -0,0 +1,43 @@
+import frc.robot.utils.CRTGearRatioUtil;
+import org.junit.jupiter.api.Test;
+import java.util.Optional;
+
+public class CRTGearRatioUtilTestingStuff {
+  @Test
+  void test() {
+    assert checker(44, 42, 22);
+    assert checker(44, 26, 22);
```
balghane, line 11: Check symmetry, e.g. does it also work for `42, 44, 22`?

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: I would need to cast the doubles to longs and then cast back, do we want to do this?

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: I made it private. Is gcd being greatest common divisor not common knowledge? I'll add a comment, I guess.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: I was being unit tested before, I can remove the unit test and make it private.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: I asked Victor about this and he said he would try to explain it more to me at some point this week. I'll put that into a variable.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: In this case it shouldn't matter as the way Victor wrote it makes it always be an integer anyway. The int cast it just to make the return type correct.

src/test/java/CRTGearRatioUtilTestingStuff.java
voidSevenSevenSix: Fixed.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: NOTE TO SELF: FIX THIS LATER

src/test/java/CRTGearRatioUtilTestingStuff.java
voidSevenSevenSix: Isn't the idea of a unit test to not check if a function is equal to itself?

src/test/java/CRTGearRatioUtilTestingStuff.java
voidSevenSevenSix: It doesn't, which is correct behavior.

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: See the comment I added

src/test/java/CRTGearRatioUtilTestingStuff.java
balghane: It's to ensure it does what you think you should be doing, which you can't do by using the same code? Because then who's watching the watcher

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: Elevator position is used later so might as well use elevatorpos with mainGearCircumference

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: the reason it was needed above was because numbers like 0.99 would output numbers that rounded to the wrong number of full rotations. here less precision is needed

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: done

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: done

src/test/java/CRTGearRatioUtilTestingStuff.java
voidSevenSevenSix: done

src/test/java/CRTGearRatioUtilTestingStuff.java
voidSevenSevenSix: done

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: done

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: fixed

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: fixed

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: done

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: done

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
voidSevenSevenSix: fixed

src/test/java/CRTGearRatioUtilTestingStuff.java
voidSevenSevenSix: fixed

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
balghane: Ah no this is fine

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
balghane: All acronyms should be specified. My work coding standard actually just straight up bans acronyms but I think this is fine with a comment

src/main/java/frc/robot/utils/CRTGearRatioUtil.java
balghane: Ah fine if unit tested. Either way

src/test/java/PoseBufferTest.java
```diff
@@ -1,4 +1,5 @@
-import edu.wpi.first.math.geometry.Pose2d;
+//TODO comment this back in
+/*import edu.wpi.first.math.geometry.Pose2d;
```
balghane: Comment this back in and upgrade your wpilib if you haven't already, should work fine

balghane (approved):

PR #13, opened 2024-02-03
Shooter cleanup

balghane:
Various shooter cleanup:
- Added functions for latching
- Added check shooter speed function
- Added shooter mode
- More comments, moved around function constants cals
- Initialize pivot abs encoder to read degrees


avakimc (approved):
i think this should be approved

PR #15, opened 2024-02-03
Intake cleanup

balghane:
Various cleanup including:
- called moveToPos in periodic
- Added comments
- Moved some things around


avakimc (approved):
approved

PR #23, opened 2024-02-08
Add Interrupt Handling Feature to the Command Groups

avakimc:
- creates a StopMostThings.java class that brings the robot to the "home" state (stops the conveyor, the shooter, and the intake rollers; brings the intake to STOWED and elevator to HOME)
- adds an "interruptible" version of each command group
          - if the command group is a "prep," then interrupting the command will bring the robot to the "home" state
          - if the command group is not a "prep," then interrupting the command will bring the robot to the command's interruptible "prep" state

### DIFF
--- a/src/main/java/frc/robot/commands/AmpPrepScore.java
+++ b/src/main/java/frc/robot/commands/AmpPrepScore.java
@@ -18,7 +18,7 @@ import frc.robot.subsystems.shooter.Shooter;
  */
 public class AmpPrepScore extends SequentialCommandGroup {
   /** Creates a new AmpPreScore. */
-  public AmpPrepScore(Elevator elevator, Intake intake) {
+  public AmpPrepScore(Intake intake, Elevator elevator, Conveyor conveyor, Shooter shooter) {
     addRequirements(elevator, intake);
 
     SequentialCommandGroup moveWhenNotSafe =
@@ -41,9 +41,13 @@ public class AmpPrepScore extends SequentialCommandGroup {
             }));
   }
 
-  /** runs AmpPrepScore but adds the layer that if the robot is interrupted, then everything is brought to the home state */
-  public static Command interruptibleAmpPrepScore(Intake intake, Elevator elevator, Conveyor conveyor, Shooter shooter) {
-    return new AmpPrepScore(elevator, intake)
+  /**
+   * runs AmpPrepScore but adds the layer that if the robot is interrupted, then everything is
+   * brought to the home state
+   */
+  public static Command interruptibleAmpPrepScore(
+      Intake intake, Elevator elevator, Conveyor conveyor, Shooter shooter) {
+    return new AmpPrepScore(intake, elevator, conveyor, shooter)
         .finallyDo(
             (boolean interrupted) -> {
               new StopMostThings(intake, elevator, conveyor, shooter);
--- a/src/main/java/frc/robot/commands/AmpPrepScore.java
+++ b/src/main/java/frc/robot/commands/AmpPrepScore.java
@@ -1,5 +1,6 @@
 package frc.robot.commands;
 
+import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.ConditionalCommand;
 import edu.wpi.first.wpilibj2.command.InstantCommand;
 import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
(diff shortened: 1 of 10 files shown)


src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -156,7 +156,7 @@ public static Command receive(Conveyor conveyor) {
   }
 
   /** Stop the conveor rollers. */
-  private static Command stop(Conveyor conveyor) {
+  public static Command stop(Conveyor conveyor) {
```
itsmesp8343, line 159: Wait why should this be public now?

avakimc: before i added interrupt handling, Conveyor.stop() was only needed within the Conveyor class because the Conveyor's individual commands to receive and whatever called stop() on their own. Now that I need to stop the conveyor in StopMostThings, I need to make stop() public

src/main/java/frc/robot/commands/SpeakerShootSequence.java
```diff
@@ -22,4 +25,13 @@ public SpeakerShootSequence(Conveyor conveyor, Shooter shooter) {
         Conveyor.shoot(conveyor),
         new InstantCommand(() -> shooter.setShooterMode(ShooterMode.IDLE)));
   }
+
+  /** runs SpeakerShootSequence but adds the layer that if the robot is interrupted, then everything is brought to the prep state */
+  public static Command interruptibleSpeakerShootSequence(Intake intake, Elevator elevator, Conveyor conveyor, Shooter shooter) {
+    return new SpeakerShootSequence(intake, elevator, conveyor, shooter)
+        .finallyDo(
+            (boolean interrupted) -> {
+              SpeakerPrepScoreSequence.interruptibleSpeakerPrepScoreSequence(intake, elevator, conveyor, shooter);
```
balghane, line 34: (a) This will mean it will always go to prep score once we've scored, not just when interrupted
(b) If shooting was interrupted, then scheduling another command on top of whatever was doing the interrupting might not be a good idea?
(b) Constructing a command doesn't start it, you'd have to construct a command then do `.schedule()` on it

src/main/java/frc/robot/commands/AmpPrepScore.java
```diff
@@ -38,4 +40,17 @@ public AmpPrepScore(Elevator elevator, Conveyor conveyer, Intake intake) {
...
+   * runs AmpPrepScore but adds the layer that if the robot is interrupted, then everything is
+   * brought to the home state
+   */
+  public static Command interruptibleAmpPrepScore(
+      Intake intake, Elevator elevator, Conveyor conveyor, Shooter shooter) {
+    return new AmpPrepScore(intake, elevator, conveyor, shooter)
+        .finallyDo(
+            (boolean interrupted) -> {
+              new StopMostThings(intake, elevator, conveyor, shooter);
+            });
```
balghane, line 54: This runs whether the command was interrupted or not, so it'll just immediately go home once it's done prepping score

src/main/java/frc/robot/commands/SpeakerShootSequence.java
balghane: I have similar comments for most of the changes in this PR

src/main/java/frc/robot/commands/GoHomeSequence.java
```diff
@@ -4,17 +4,19 @@
...
 import frc.robot.subsystems.intake.Intake.IntakePosition;
+import frc.robot.subsystems.shooter.Shooter;
 
 /**
  * brings the elevator and intake to their respective home positions in a "smart" way so they do not
  * end up attacking each other accidentally
  */
 public class GoHomeSequence extends SequentialCommandGroup {
-  public GoHomeSequence(Intake intake, Elevator elevator) {
+  public GoHomeSequence(Intake intake, Elevator elevator, Conveyor conveyor, Shooter shooter) {
```
balghane, line 19: Did you intend to add these as reqs or no?

balghane (changes requested):

src/main/java/frc/robot/commands/GoHomeSequence.java
avakimc: i thought we needed to command everything ?

PR #27, opened 2024-02-10
Add intake limelight

tails618:
- Mostly a copy from last year


tails618:
don't review yet it's not done lol

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+
+    /**
+     * We convert to 2d target space: X+ -> Out of the target Y+ -> Pointing to the right of the
+     * target (If you are looking at the target) This means positive yaw is based on Z+ being up
+     */
+    Translation2d translation =
+        new Translation2d(botPoseTargetSpace.getZ(), botPoseTargetSpace.getX());
+    // Rotation2d rot = Rotation2d.fromDegrees(-botPoseTargetSpace.getRotation().getY());
+
+    Rotation2d rot = Rotation2d.fromDegrees(0);
```
balghane: Let's maybe return the real rotation from the tag for this year. Last year, we knew every tag was going to be at 0 deg but this year, I'm concerned about assuming anything about how the stage is placed on the field. So we may want to use the actual rotation returned by the tag detection thingy

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+  }
+
+  private static boolean validScoringTag(double tagId) {
+    long tagIdRounded = Math.round(tagId);
+    if (tagIdRounded == 1
+        || tagIdRounded == 2
+        || tagIdRounded == 3
+        || tagIdRounded == 6
+        || tagIdRounded == 7
+        || tagIdRounded == 8) {
```
balghane: Intake cam only cares about the six stage tags, can you update for that?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+        bestTag = tagIndex;
+      }
+    }
+    return bestTag;
+  }
+
+  private Transform2d getRobotToScoringLocation(Pose3d targetPoseRobotSpace) {
+    Transform2d targetFromBot = getBotFromTarget(targetPoseRobotSpace);
+    return targetFromBot;
+  }
```
balghane, line 185: Do we need this? What is it doing?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+
+  public Optional<Transform2d> getRobotToScoringLocation() {
+    if (getPipeline() != pipeline.PIPELINE1) {
+      setPipeline(pipeline.PIPELINE1);
+    }
+    return robotToScoringLocation;
+  }
```
balghane: Do we need this? What is it doing?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+    }
+    robotToScoringLocation =
+        Optional.of(getRobotToScoringLocation(LimelightHelpers.getTargetPose3d_RobotSpace("")));
+    return robotToScoringLocation;
+  }
+
+  public double getLatencySeconds() {
+    if (getPipeline() != pipeline.PIPELINE1) {
+      setPipeline(pipeline.PIPELINE1);
+    }
```
balghane: This will be useful for either tags or notes

balghane: Can you dig through the docs on whether this correctly captures the latency for both neural network and tag detection pipelines?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+  //AprilTag detection
+  private Optional<Transform2d> robotToScoringLocation = Optional.empty();
+
+
+
+  /**
+   * Create an IntakeLimelight object
+   *
+   * @param angleDegrees angle from normal in degress. Looking straight out is 0, and increasing as
+   *     the camera is tilted towards the ceiling.
```
balghane: This is describing pitch

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+   * in the constructor.
+   *
+   * @param targetHeight height of the target in meters
+   * @return distance to the target in meters
+   */
+  public double getDistanceFromTargetMeters(double targetHeight) {
+    double angle = kCameraAngleDegrees + getOffSetY();
+    if (angle < 1 || angle > 89) {
+      return 0;
+    }
```
balghane: Why is <1 not valid here?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+   * Create an IntakeLimelight object
+   *
+   * @param angleDegrees angle from normal in degress. Looking straight out is 0, and increasing as
+   *     the camera is tilted towards the ceiling.
+   * @param heightMeters height of the camera measured from the lens to the ground in meters.
+   * @param targetHeightMeters height to the center of the target in meters
+   */
+  public IntakeLimelight(double angleDegrees, double heightMeters, double targetHeightMeters) {
+    kCameraAngleDegrees = angleDegrees;
+    kCameraHeight = heightMeters;
```
balghane, line 75: This is not fixed, since the camera is on the elevator. We should pass it in when we care about it

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+   * @param targetHeight height of the target in meters
+   * @return distance to the target in meters
+   */
+  public double getDistanceFromTargetMeters(double targetHeight) {
+    double angle = kCameraAngleDegrees + getOffSetY();
+    if (angle < 1 || angle > 89) {
+      return 0;
+    }
+    double tanTerm = Math.tan(Math.toRadians(angle));
+    double distance = (targetHeight - kCameraHeight) / tanTerm;
```
balghane: Shouldn't this be `(targetHeight - kCameraHeight) * tanTerm;`?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+    y_norm = zy_translation.getY();
+
+    /**
+     * Find the intersection between the target in space, and the vector pointing to the target
+     *
+     * <p>This becomes a vector [x_norm, y_norm, z_norm] * d = [x_target, diff_height, z_target]
+     *
+     * <p>x_norm * d = x_target y_norm * d = diff_height z_target * d = z_target
+     */
+    double scaling = diff_height / y_norm;
```
balghane, line 415: Potential for a divide by zero here

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+  }
+
+  public class NoteDetection {
+    public double latencySec;
+
+    /** Distance from camera */
+    public double distanceMeters;
+
+    /** CCW Angle */
+    public double angleDeg;
```
balghane: Is this pitch or yaw

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+  }
+
+  /**
+   * Looks for a note
+   *
+   * @return If empty, no note detected. First value is note yaw in degrees (ccw), second value is
+   *     distance in meters.
+   */
+  public Optional<NoteDetection> getNotePos() {
+    if (getPipeline() != pipeline.PIPELINE0) {
```
balghane: Can we note somewhere above that 0 is Note and 1 is Tag

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+   * https://www.chiefdelphi.com/t/what-does-limelight-skew-actually-measure/381167/7
+   *
+   * @return Translation2d from the camera to the target. This is camera relative, x is out of the
+   *     camera and positive y is to the left from the camera's point of view.
+   */
+  public Translation2d getTargetTranslation() {
+    return getTargetTranslation(kTargetHeight);
+  }
+
+  public Optional<Double> getXOfSmallestY(double[] corners) {
```
balghane: Can you add a comment that this grabs the object that is lowest in the image

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+   *     distance in meters.
+   */
+  public Optional<NoteDetection> getNotePos() {
+    if (getPipeline() != pipeline.PIPELINE0) {
+      setPipeline(pipeline.PIPELINE0);
+    }
+
+    double[] corners = table.getEntry("tcornxy").getDoubleArray(new double[0]);
+    Optional<Double> maybeXPixels = getXOfSmallestY(corners);
+    Optional<Double> maybeYHeightPixels = getObjectHeightPx(corners);
```
balghane: Do we need to filter out low confidence detections?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+
+    if (!maybeXPixels.isPresent() || !maybeYHeightPixels.isPresent() || table.getEntry("tclass").getString("") != "note") {
```
balghane: `table.getEntry("tclass")` This is probably only the class for the object that the limelight selected, isn't it?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+   *
+   * @return If empty, no note detected. First value is note yaw in degrees (ccw), second value is
+   *     distance in meters.
+   */
+  public Optional<NoteDetection> getNotePos() {
+    if (getPipeline() != pipeline.PIPELINE0) {
+      setPipeline(pipeline.PIPELINE0);
+    }
+
+    double[] corners = table.getEntry("tcornxy").getDoubleArray(new double[0]);
```
balghane: Does this include `robot` and `note`?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+    // Compute note distance
+    // kinda based on the same thing as below:
+    // https://docs.limelightvision.io/en/latest/theory.html#from-pixels-to-angles
+    final double vertFovDeg = 49.7;
+    double halfResYPixels = RESOLUTION_Y / 2.0;
+    double viewplaneHeightPixels = 2.0 * Math.tan(Units.degreesToRadians(vertFovDeg / 2));
+    double yPixels = maybeYHeightPixels.get();
+    double normYPixels = (1 / halfResYPixels) * yPixels;
+    double viewplaneYPixels = viewplaneHeightPixels / 2.0 * normYPixels;
+    final double CONE_HEIGHT_METERS = Units.inchesToMeters(13.0);
```
balghane: Update

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,610 @@
...
+    double normXPixels = (1 / halfResXPixels) * ((halfResXPixels - 0.5) - xPixels);
+    double viewplaneWidthPixels = 2.0 * Math.tan(Units.degreesToRadians(horizFovDeg / 2));
+    double viewplaneXPixels = viewplaneWidthPixels / 2.0 * normXPixels;
+    double angleXDegrees = Units.radiansToDegrees(Math.atan2(viewplaneXPixels, 1));
+    double angleAdjustDegrees = -3.0; // due to limelight yaw
+    double adjustedAngleDegrees = angleXDegrees + angleAdjustDegrees;
+    return Optional.of(new NoteDetection(getLatency(), coneDistanceMeters, adjustedAngleDegrees));
+  }
+
+  public Optional<Double> getAngleToNoteDeg() {
```
balghane: Do we need this if we have the above?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightConstants.java
```diff
@@ -0,0 +1,5 @@
+package frc.robot.subsystems.intakeLimelight;
+
+public class IntakeLimelightConstants {
+
```
balghane: Camera pitch goes here probably

balghane (changes requested):

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Isn't that what it should be?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: This is the intake Limelight, which I believe is on the conveyer (that's what it looked like in [this image](https://media.discordapp.net/attachments/673346755072491521/1201927233355124807/Screenshot_2024-01-30_102653.png?ex=65de0df7&is=65cb98f7&hm=e1026de3d70d36a230912d0a29fb25e89f532bdb3262f53603ffcebb7a844b52&=&format=webp&quality=lossless&width=1640&height=1194)), so the height should be constant (the conveyer should always be in the same place when intaking)

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: It's in the comment on `pipeline`

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Fixed. I *could* check our math state to isolate only our alliance's stage but I don't think that's necessary; whenever we're using this, it should be obvious which tag to look at

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Used in `checkForTag`, which is used in `LookForTag` (which doesn't exist here yet but was used last year). It seems like it gets a transform from the robot to the tag (scoring location).

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Returns the existing variable itself, set in a few different places. This function is public, whereas the above function is private.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Changed

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Don't see why it wouldn't, pretty sure everything in [this section of the LL NetworkTables API](https://docs.limelightvision.io/docs/docs-limelight/apis/complete-networktables-api#basic-targeting-data) applies to all pipelines.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: No idea. Removed.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: No. Whiteboard math here: 
photo
I can type out an explanation if it's unclear.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: I don't think I grasp the math well enough to figure out how to avoid it... can I get a hint?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Done

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Yes, I believe if we aren't explicitly limiting based on class it will look for everything in the model (unless there's also filtering which classes it detects within the limelight console itself)

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: I don't see anything about that in the API; I assume/hope it can be done in the LL dashboard.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: It is, but I think it's the right thing to do; there's other stuff here that's also only focused on the target (like `corners`, I think). 

We could get the JSON dump of *every* object it detects in the frame; do you think that's necessary?

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Done

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Nope! Removed.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Yaw

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightConstants.java
tails618: Done. I think it's 0.0deg, correct me if I'm wrong

tails618 (commented):
Responses to comments

tails618 (commented):
Responses to comments

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: Comment should say it's pitch then

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: Hmmm okay

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: Generally you can just check if it's within some epsilon of zero and if yes, then replace it with a very small but non-zero value (like 1e-4 or something)

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: ok let's remember to test that

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: corners is the corners for all the objects, so hopefully there's a class for all the objects too?

balghane: Can you put this on the minibot and test?

balghane: Here's a model you can throw on the minibot https://www.chiefdelphi.com/t/limelight-neural-network-note-detection/449822/10?u=basel_a

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: Could be an issue, we only want to see notes here...

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightConstants.java
balghane: I don't know, didn't see it in the sheet

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,575 @@
...
+    // Compute note distance
+    // kinda based on the same thing as below:
+    // https://docs.limelightvision.io/en/latest/theory.html#from-pixels-to-angles
+    final double vertFovDeg = 49.7;
+    double halfResYPixels = RESOLUTION_Y / 2.0;
+    double viewplaneHeightPixels = 2.0 * Math.tan(Units.degreesToRadians(vertFovDeg / 2));
+    double yPixels = maybeYHeightPixels.get();
+    double normYPixels = (1 / halfResYPixels) * yPixels;
+    double viewplaneYPixels = viewplaneHeightPixels / 2.0 * normYPixels;
+    final double NOTE_HEIGHT_METERS = Units.inchesToMeters(13.0);
```
balghane: A note is not 13 inches tall lol. It's just above the ground, we can call it 1 inch

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Done

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Should you check if it is ACTUALLY 0, and if not replace it with a very small but non-zero value? We don't care if it's within some epsilon of 0, that won't cause a divide by 0 issue.

Fixed based on this, but if it should actually be checking if it's within some epsilon I'll change it

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Ah, thought you were asking about that line specifically. In this function we check if it's a note on L535:

```java
if (!maybeXPixels.isPresent() || !maybeYHeightPixels.isPresent() || table.getEntry("tclass").getString(IntakeLimelightConstants.INTAKE_LIMELIGHT_NAME) != "note") {
      System.out.println("Didn't see note");
      return Optional.empty();
    }
```

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: What if it's balanced *perfectly* upright and all the teams agree to not move their robots so it doesn't fall?? This feels like a very realistic scenario and would result in a 14" tall note.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Yeah, we'll test tomorrow. I'm leaving in a few minutes so I'm not gonna work on it rn.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightConstants.java
tails618: Nick wants us to add Limelight stuff to the sheet so he can go through and add that stuff. for know I've made it a placeholder.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: `table.getEntry("tclass")` this almost certainly only checks if the "limelight selected object" is a note, not the one we have selected above which could be either

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: Even if it's 1e-7 that's bad

tails618: What do you mean? Even if `y_norm` is 1e-7? I don't see why that would be an issue, it should divide fine (it'll mean `scaling` is giant, but it would also be giant if I did 1e-4 like you suggested earlier).

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
balghane: Doesn't look removed to me

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,579 @@
...
+   *
+   * @return Translation2d from the camera to the target. This is camera relative, x is out of the
+   *     camera and positive y is to the left from the camera's point of view.
+   */
+  public Translation2d getTargetTranslation() {
+    return getTargetTranslation(kTargetHeight);
+  }
+
+  /** Gets the object lowest in the image */
+  public Optional<Double> getXOfSmallestY(double[] corners) {
```
balghane: Not using this anymore

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,579 @@
...
+      if (corners[i] < minY) {
+        minY = corners[i];
+        minYIdx = i;
+      }
+    }
+
+    return Optional.of(corners[minYIdx - 1]);
+  }
+
+  public Optional<Double> getObjectHeightPx(double[] corners) {
```
balghane: Not using this

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,579 @@
...
+   *     distance in meters.
+   */
+  public Optional<NoteDetection> getNotePos() {
+    // TODO filter low confidence detection in LL dashboard, hopefully
+    if (getPipeline() != pipeline.NOTE_PIPELINE) {
+      setPipeline(pipeline.NOTE_PIPELINE);
+    }
+
+    LimelightHelpers.LimelightResults llresults =
+        LimelightHelpers.getLatestResults(IntakeLimelightConstants.INTAKE_LIMELIGHT_NAME);
```
balghane, line 472: There was something about the first time you run this, it takes a couple seconds? Something with the json parsing library. So you should make sure to do it before the match.

Not sure if doing it in the constructor works. Maybe disabledInit?

https://www.chiefdelphi.com/t/introducing-limelight-lib/425660/57?u=basel_a

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,579 @@
...
+    for (LimelightTarget_Detector detection : targets_Detector) {
+      if (detection.className == "note") {
+        if (detection.ty_pixels > lowestDetection.ty_pixels) { // lower in image = greater y value
+          lowestDetection = detection;
+        }
+      }
+    }
+
+    double angleLimelightToNote =
+        IntakeLimelightConstants.INTAKE_LIMELIGHT_PITCH_DEGREES + lowestDetection.ty;
```
balghane: I think `INTAKE_LIMELIGHT_PITCH_DEGREES` is defined as 0 parallel to ground, 90 up. How is ty defined?

balghane: Add units to var name also

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,579 @@
...
+    builder.addDoubleProperty("Target Area", () -> getTargetArea(), null);
+    builder.addDoubleProperty("Skew", () -> getSkew(), null);
+    builder.addDoubleProperty("Latency", () -> getLatency(), null);
+    builder.addDoubleProperty("Distance", () -> getDistanceFromTargetMeters(), null);
+    builder.addDoubleProperty("Tx", () -> getOffSetX(), null);
+    builder.addDoubleProperty("Ty", () -> getOffSetY(), null);
+    builder.addDoubleProperty("Translation Distance", () -> m_lastDistance, null);
+    builder.addDoubleProperty("Translation X", () -> m_lastX, null);
+    builder.addDoubleProperty("Translation Y", () -> m_lastY, null);
+    builder.addBooleanProperty("Valid Target", () -> isValidTarget(), null);
```
balghane: We might want to remove most of this? Or anything we're not intending on using initially

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightCal.java
```diff
@@ -0,0 +1,5 @@
+package frc.robot.subsystems.intakeLimelight;
+
+public class IntakeLimelightCal {
+    public static final double LIMELIGHT_YAW = 0.0;
```
balghane: add units

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightConstants.java
```diff
@@ -0,0 +1,10 @@
+package frc.robot.subsystems.intakeLimelight;
+
+import frc.robot.Constants;
+
+public class IntakeLimelightConstants {
+  public static final String INTAKE_LIMELIGHT_NAME = "intake";
```
balghane: I think the name is always `limelight-XYZ` so `limelight-intake`?

balghane (changes requested):

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: ok well nothing's perfect (except now i've actually removed it so it is perfect)

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Removed

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Removed

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: Added to `disabledInit`

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: "Y-coordinate of the center of the target in degrees. Positive-down, center-zero" so 0 is fine, and sign is also correct. (If we have a note 10deg below horizontal, that is +10 in limelight degrees, and if our limelight is angled 10 degrees up, that is +10 in robot degrees; (+10) + (+10) = +20, which I believe is what we want.

I have changed the name of the var.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
tails618: I removed some stuff we definitely aren't using.  Left some stuff that might be useful.

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightCal.java
tails618: Done

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelightConstants.java
tails618: Good catch

src/main/java/frc/robot/subsystems/intakeLimelight/IntakeLimelight.java
```diff
@@ -0,0 +1,511 @@
...
+    for (LimelightTarget_Detector detection : targets_Detector) {
+      if (detection.className == "note") {
+        if (detection.ty_pixels > lowestDetection.ty_pixels) { // lower in image = greater y value
+          lowestDetection = detection;
+        }
+      }
+    }
+
+    double angleLimelightToNoteDegrees =
+        IntakeLimelightConstants.INTAKE_LIMELIGHT_PITCH_DEGREES + lowestDetection.ty;
```
balghane: The way we're using this below is that it's the angle relative to field horizontal, from the limelight position. So if the result (note relative to field horizontal) is 4 deg down and the limelight is 7 deg up. Then the ty will report 11. So I think we need `ty - INTAKE_LIMELIGHT_PITCH_DEGREES `

tails618: Yup, you're right. Fixed

balghane (approved):

PR #28, opened 2024-02-10
Light Class

allison324:
We made Light class and mainly copied everything from last year's code(same functions and everything) - the only things that we changed are the colors and the enums. Also added a const int for the canid in RobotMap.java.


src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+import frc.robot.RobotMap;
+import java.util.TreeMap;
+
+public class Lights {
+  private TreeMap<LightCode, Integer[]> lightOptionsMap;
+  private CANdle candle = new CANdle(RobotMap.CANDLE_CAN_ID);
+  private CANdleConfiguration config = new CANdleConfiguration();
+  private LightCode currentLightStatus = LightCode.OFF;
+
+  public enum LightCode{
```
avakimc: probably add a HOLDING_NOTE enum

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+    NOTELESS, //red
+    INTAKING, //blue
+    READY_TO_SHOOT, //green
+    NO_TAG, //purple
+    OFF, //grey
+    ALIGNING_TO_TAG, //orange
+    PARTY_MODE;
```
avakimc: I would reorder these so that OFF and PARTY_MODE are at the end. this is nitpicky though

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+  private CANdle candle = new CANdle(RobotMap.CANDLE_CAN_ID);
+  private CANdleConfiguration config = new CANdleConfiguration();
+  private LightCode currentLightStatus = LightCode.OFF;
+
+  public enum LightCode{
+    NOTELESS, //red
+    INTAKING, //blue
+    READY_TO_SHOOT, //green
+    NO_TAG, //purple
+    OFF, //grey
```
avakimc: I could be wrong, but shouldn't OFF just be no lights are on?

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+    candle.configAllSettings(config);
+
+    //array of integers that represents the R, G, B values of each lightcode enum
+    lightOptionsMap = new TreeMap<LightCode, Integer[]>();
+    lightOptionsMap.put(LightCode.NOTELESS, new Integer[] {255, 0, 0});
+    lightOptionsMap.put(LightCode.INTAKING, new Integer[] {0, 0, 255});
+    lightOptionsMap.put(LightCode.ALIGNING_TO_TAG, new Integer[] {0, 255, 0});
+    lightOptionsMap.put(LightCode.NO_TAG, new Integer[] {255, 0, 255});
+    lightOptionsMap.put(LightCode.OFF, new Integer[] {120, 120, 120});
+    lightOptionsMap.put(LightCode.ALIGNING_TO_TAG, new Integer[] {255, 165, 0});
```
avakimc: 1) ALIGNING_TO_TAG is in here twice --> copy-paste typo?
2) I think that the first ALIGNING_TO_TAG is supposed to be READY_TO_SHOOT because you say that is the green color earlier

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  private void setNoTag() {
+    StrobeAnimation strobeAnim = new StrobeAnimation(0, 255, 0);
+    candle.animate(strobeAnim);
+  }
+
+  public void setPartyMode() {
+      RainbowAnimation rainbowAnim = new RainbowAnimation(1, 0.5, LightsConstants.NUM_CANDLE_LEDS);
```
avakimc: i don't know if this is necessary, but usually we make variables for each value so they aren't unlabeled. so maybe make variables in LightsConstatnts.java for the rest of these parameters (1, 0.5)?

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+      currentLightStatus = light;
+    }
+    setLEDs();
+  }
+
+  private void setLEDs() {
+    if (currentLightStatus == LightCode.PARTY_MODE) {
+      setPartyMode();
+    } else if (currentLightStatus == LightCode.NO_TAG) {
+      setNoTag();
```
avakimc, line 60: I know you were looking at the minibot Lights class for inspiration. The reason why we needed this second if-statement in the minibot was because NO_TAG was a blinking strobe animation (like how PARTY_MODE is an animation).

if NO_TAG is just a solid color like the other LightCode options (which is what you have commented and in the tree map) then you don't need a separate if statement and function to change the lights to the NO_TAG colorway

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,78 @@
...
+        lightOptionsMap.get(currentLightStatus)[0],
+        lightOptionsMap.get(currentLightStatus)[1],
+        lightOptionsMap.get(currentLightStatus)[2]);
+    }
+  }
+
+  private void setNoTag() {
+    StrobeAnimation strobeAnim = new StrobeAnimation(0, 255, 0);
+    candle.animate(strobeAnim);
+  }
```
avakimc: this isn't needed if you're making NO_TAG a solid color (which is what you have defined in the tree map). We can have a real discussion about the colors/patterns we want to use later on though.

src/main/java/frc/robot/RobotMap.java
```diff
@@ -31,4 +31,8 @@ public class RobotMap {
   public static final int INTAKE_PIVOT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
       INTAKING_FRONT_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT,
       INTAKING_BACK_MOTOR_CAN_ID = Constants.PLACEHOLDER_INT;
+
+    /** CANdle for the lights */
```
avakimc: this is nitpicky but the comment is kind of misleading, in my opinion. either delete it or edit

avakimc (changes requested):
i have requested some changes for this class

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,80 @@
...
+    OFF, //grey
+    PARTY_MODE;
+  }
+
+  public Lights() {
+    config.stripType = LEDStripType.RGB;
+    config.brightnessScalar = 1.0;
+    candle.configAllSettings(config);
+
+    //array of integers that represents the R, G, B values of each lightcode enum
```
avakimc: this comment is not accurate as lightOptionsMap is not an array. also i would use the /** comment here */ format for the comment so that hovering over the variable lightOptionsMap will display the comment. make sure to put that comment on the line where you say private TreeMap<LightCode, Integer[]> lightOptionsMap; (line 14) though.

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,80 @@
...
+  private LightCode currentLightStatus = LightCode.OFF;
+
+  public enum LightCode{
+    NOTELESS, //red
+    INTAKING, //blue
+    READY_TO_SHOOT, //green
+    NO_TAG, //blinking purple
+    ALIGNING_TO_TAG, //orange
+    HOLDING_NOTE, //yellow
+    OFF, //grey
```
avakimc: if OFF is not 0,0,0 the comment shouldn't be "grey" right? shouldn't it be like, "no lights are on" or something?

src/main/java/frc/robot/subsystems/lights/LightsConstants.java
```diff
@@ -0,0 +1,9 @@
+package frc.robot.subsystems.lights;
+
+import frc.robot.Constants;
+
+public class LightsConstants {
+  public static final int NUM_CANDLE_LEDS = Constants.PLACEHOLDER_INT;
+  public static final int LIGHT_BRIGHTNESS = 1;
```
avakimc: I could be wrong here but I looked over the documentation and "brightness" is also a double. therefore, LIGHT_BRIGHTNESS should be a double with a value of 1.0. I think that having it as an int also works but generally we like to make it a double and add the .0 to be more clear
Screenshot 2024-02-12 at 7 20 50 PM

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,80 @@
...
+        lightOptionsMap.get(currentLightStatus)[0],
+        lightOptionsMap.get(currentLightStatus)[1],
+        lightOptionsMap.get(currentLightStatus)[2]);
+    }
+  }
+
+  private void setNoTag() {
+    StrobeAnimation strobeAnim = new StrobeAnimation(255, 0, 255);
+    candle.animate(strobeAnim);
+  }
```
avakimc, line 72: you still have NO_TAG as a blinking animation and as a solid color (from the treemap). you shouldn't / can't have both

avakimc (changes requested):
added some more comments

src/main/java/frc/robot/subsystems/lights/Lights.java
tails618: Agree that we want the comment to display when hovering, but that means the comment must be above where we *create* the variable, on line 14.

src/main/java/frc/robot/subsystems/lights/LightsConstants.java
tails618: Yes, brightness should be a double for animations (it's [0,1], so if it was an int it wouldn't really be brightness and it would just be on/off). An integer will work (Java can typecast an integer n to a double n.0) but it lacks clarity and there really isn't a reason to have it as an int here so this should be a double.

src/main/java/frc/robot/subsystems/lights/Lights.java
```diff
@@ -0,0 +1,80 @@
...
+    } else {
+      currentLightStatus = light;
+    }
+    setLEDs();
+  }
+
+  private void setLEDs() {
+    if (currentLightStatus == LightCode.PARTY_MODE) {
+      setPartyMode();
+    } else if (currentLightStatus == LightCode.NO_TAG) {
```
avakimc, line 59: you deleted NO_TAG from the treemap, which was good. you also deleted it as a LightCode but you still need NO_TAG as an enum option (that way you can set it as a desired color and stuff). so i don't think this will build

avakimc (changes requested):
last requested change

avakimc (approved):
approved. probably don't merge until basel or asher also approve

tails618 (approved):

PR #30, opened 2024-02-15
Make sure elevator and shooter don't intersect

balghane:
This mostly just ensures the shooter doesn't hit the conveyor when the elevator moves.

Also:
- one bug in IntakeSequence (below zone vs above zone)
- fixed intake using offset encoder position instead of directly using encoder position
- Removed some example stuff


src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -26,6 +26,8 @@ public class Shooter extends SubsystemBase {
   public enum ShooterMode {
     /** Not doing anything */
     IDLE,
+    /** Spin up the shooter but stay at home position */
+    SPIN_UP,
     /** Shooting or spinning up to shoot */
     SHOOT,
```
tails618: If `SPIN_UP` is for spinning up but staying at home position, modify the comment on `SHOOT` to reflect that the position is different

src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -228,6 +234,10 @@ public void periodic() {
         stopShooter();
         controlPosition(ShooterCal.STARTING_POSITION_DEGREES);
         break;
+      case SPIN_UP:
+        spinUpShooter();
+        controlPosition(ShooterCal.STARTING_POSITION_DEGREES);
```
tails618, line 239: Do we need to check that everything is out of the danger zone?

src/main/java/frc/robot/commands/IntakeSequence.java
```diff
@@ -9,26 +9,34 @@
...
             intake),
+        new InstantCommand(
+            () -> {
+              shooter.setShooterMode(ShooterMode.IDLE);
+            }),
         new ConditionalCommand(
             new InstantCommand(),
-            new WaitUntilCommand(intake::clearOfConveyorZone),
-            elevator::elevatorAboveInterferenceZone),
+            new SequentialCommandGroup(
```
tails618, line 36: This should maybe be parallel? I don't remember exactly which type it is to wait for them both to finish. I suppose functionally it doesn't really matter but it does make it a little clearer IMO

tails618: I feel this way wherever we have consecutive waits in a sequential command group. It doesn't *really* matter but they shouldn't be thought of as "wait for one then wait for the other" which is what a sequential command group suggests.

src/main/java/frc/robot/commands/AmpPrepScore.java
```diff
@@ -9,21 +9,24 @@
...
+  public AmpPrepScore(Elevator elevator, Conveyor conveyer, Intake intake, Shooter shooter) {
+    addRequirements(conveyer, elevator, intake, shooter);
 
     SequentialCommandGroup moveWhenNotSafe =
         new SequentialCommandGroup(
             new InstantCommand(
                 () -> intake.setDesiredIntakePosition(IntakePosition.CLEAR_OF_CONVEYOR)),
+            new InstantCommand(() -> shooter.setShooterMode(Shooter.ShooterMode.IDLE)),
             new WaitUntilCommand(intake::clearOfConveyorZone),
+            new WaitUntilCommand(shooter::clearOfConveyorZone),
```
tails618, line 29: Sequential wait commands (see intakesequence comment)

src/main/java/frc/robot/commands/ClimbPrepSequence.java
```diff
@@ -12,19 +12,22 @@
...
  */
-public class PrepClimbSequence extends SequentialCommandGroup {
-  public PrepClimbSequence(Intake intake, Elevator elevator, Shooter shooter) {
+public class ClimbPrepSequence extends SequentialCommandGroup {
+  public ClimbPrepSequence(Intake intake, Elevator elevator, Shooter shooter) {
     addRequirements(intake, elevator, shooter);
+    SequentialCommandGroup waitTillSafe =
+        new SequentialCommandGroup(
+            new WaitUntilCommand(shooter::clearOfConveyorZone),
+            new WaitUntilCommand(intake::clearOfConveyorZone));
```
tails618, line 24: I don't know if this is quite necessary to be it's own commandgroup. It works, of course, but it's pretty clear what's happening from the two `WaitUntilCommand`s (and we don't do this anywhere else - we should always do it this way or always not).

src/main/java/frc/robot/commands/GoHomeSequence.java
```diff
@@ -8,36 +8,34 @@
 import frc.robot.subsystems.elevator.Elevator.ElevatorPosition;
 import frc.robot.subsystems.intake.Intake;
 import frc.robot.subsystems.intake.Intake.IntakePosition;
+import frc.robot.subsystems.shooter.Shooter;
+import frc.robot.subsystems.shooter.Shooter.ShooterMode;
 
 /**
- * brings the elevator and intake to their respective home positions in a "smart" way so they do not
- * end up attacking each other accidentally
+ * Brings the elevator, intake, an shooter to their respective home positions in a "smart" way so
```
tails618: `an shooter` -> `and shooter` (i wouldn't request changes just for this but if you're already changing other stuff, might as well)

src/main/java/frc/robot/commands/GoHomeSequence.java
tails618, line 1: Why did we change this to make it end before everything is home?

tails618 (changes requested):

src/main/java/frc/robot/subsystems/shooter/Shooter.java
balghane: Going to start is always safe for the shooter

src/main/java/frc/robot/commands/IntakeSequence.java
balghane: This is easier to follow and the outcome is the same?

src/main/java/frc/robot/commands/ClimbPrepSequence.java
balghane: Is it worth changing?

balghane: I feel like this one is borderline but probably fine either way

src/main/java/frc/robot/commands/GoHomeSequence.java
balghane: I think it's a better practice to do the waits at the start of the commands that care, so that commands that don't care are allowed to proceed right away

src/main/java/frc/robot/commands/GoHomeSequence.java
balghane: Sure fixed

src/main/java/frc/robot/subsystems/shooter/Shooter.java
balghane: Yep changed

src/main/java/frc/robot/commands/IntakeSequence.java
tails618: I feel like parallel is a little easier to follow but it works the same and that's just my opinion so not a big deal

src/main/java/frc/robot/commands/ClimbPrepSequence.java
tails618: Fine either way so let's leave it.

tails618 (approved):
Approved

PR #37, opened 2024-02-18
Change in shooter angles while moving util

Panda8188:
* Calculate change in "azimuth" angle based on tangential velocity
* Calculate change in "elevation" angle based on radial velocity

For shooter
*TODO: Implement util in shooter code given current tangential, radial velocity, distance from speaker, and current azimuth angle (needed unfortunately). simple tan logic.

google doc is being updated as of this post and will have a detailed explanation regarding all variables and related silly logic

The name "delta angle speed calc util" may be silly goofy, but can be understood as such:
Delta = change in
Angle = angle
(based on)
speed = speed/velocity
calc = calculation
util = utility

So, one can see: The change in angle based on speed is calculated in this utility


src/main/java/frc/robot/utils/DeltaAngleSpeedCalcUtil.java
```diff
@@ -0,0 +1,46 @@
...
+      double distFromSpeakerMeters,
+      double currentAzimuthAngleDegrees) {
+
+
+    /** 
+     * :)))))))))))))))))))
+     * https://docs.google.com/document/d/1rIMmzDA1EMH7NQexIMrNYAD-_KFLe9cL4zPLXdl0lek/edit?usp=sharing
+     */
+    double calculatedAzimuthAngleDegrees = Math.atan(tangentialVelocityMPS / SHOOTER_SPEED_MPS);
+    double deltaAzimuthAngleDegrees = currentAzimuthAngleDegrees - calculatedAzimuthAngleDegrees;
```
balghane: Isn't `calculatedAzimuthAngleDegrees` itself the delta?

src/main/java/frc/robot/utils/DeltaAngleSpeedCalcUtil.java
```diff
@@ -0,0 +1,46 @@
+package frc.robot.utils;
+
+import edu.wpi.first.math.Pair;
+import edu.wpi.first.math.util.Units;
+import frc.robot.Constants;
+
+public class DeltaAngleSpeedCalcUtil {
+  private static final double SPEAKER_HEIGHT_INCHES = Units.inchesToMeters(77.9528);
+  private static final double SHOOTER_SPEED_MPS = Constants.PLACEHOLDER_DOUBLE;
```
balghane: We're going to have to measure this at some point. For now can you set it to like 10 or something?

Also it won't be possible to unit test without setting this to some non-zero value.

Can you make sure your unit test still passes if we change this number? Like maybe make this public and non-final, and set it in your unit test. Or make this variable and the below function non-static, and set this variable in the constructor. You've got different options.

Panda8188: Alright

src/main/java/frc/robot/utils/DeltaAngleSpeedCalcUtil.java
Panda8188: Actually I think you're right.. good point.

src/main/java/frc/robot/utils/DeltaAngleSpeedCalcUtil.java
```diff
@@ -0,0 +1,46 @@
+package frc.robot.utils;
+
+import edu.wpi.first.math.Pair;
+import edu.wpi.first.math.util.Units;
+import frc.robot.Constants;
+
+public class DeltaAngleSpeedCalcUtil {
+  private static final double SPEAKER_HEIGHT_INCHES = Units.inchesToMeters(77.9528);
```
avakimc: unless I'm misinterpreting, wouldn't inchesToMeters() return a value in meters? so shouldn't SPEAKER_HEIGHT_INCHES actually be SPEAKER_HEIGHT_METERS?

avakimc (commented):
small comment. i am not approving because I don't really understand the math

src/main/java/frc/robot/utils/DeltaAngleSpeedCalcUtil.java
Panda8188: Great point thanks 
I will be making the google doc much more easier to read but here's the link https://docs.google.com/document/d/1rIMmzDA1EMH7NQexIMrNYAD-_KFLe9cL4zPLXdl0lek/edit?usp=sharing
its kinda long so sorry about that

the issue is the diagrams kinda take a little to make because i have to replicate it from my own notes and sometimes my notes are illegible..

Panda8188:
ok so everything kinda died here so im going to fix it all by rebasing off of main which is wisco-cleanup.

PR #39, opened 2024-02-22
Update WPILib to 2024.3.1

tails618:


build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id "java"
-    id "edu.wpi.first.GradleRIO" version "2024.2.1"
+    id "edu.wpi.first.GradleRIO" version "2024.3.1"
```
AdityaGupta0, line 3: can we check units on this?

AdityaGupta0 (approved):

build.gradle
tails618: yeah so actually you can see it says "version" right there and it's in units of versions so it's correct I would say

AdityaGupta0: ok cool approved

PR #40, opened 2024-02-22
Shooter Limelight

tails618:
* Adds the shooter side Limelight, which only needs to detect tags.
* Code was originally copied over from intake side Limelight (with intake-specific stuff renamed)
* Removed note detection functionality, shooter side only needs to detect tags
* Moved `camMode`, `ledMode`, and `pipeline` enums from intake Limelight to Constants (and deleted them from shooter Limelight) so they can be used across the two subsystems.

### DIFF
--- a/src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
+++ b/src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
@@ -77,7 +77,7 @@ public class ShooterLimelight extends SubsystemBase {
     setLimelightValues(
         Constants.limelightLedMode.OFF,
         Constants.limelightCamMode.VISION_PROCESSING,
-        Constants.limelightPipeline.NOTE_PIPELINE);
+        Constants.limelightPipeline.TAG_PIPELINE);
 
     m_simDevice = SimDevice.create(ShooterLimelightConstants.SHOOTER_LIMELIGHT_NAME);
     if (m_simDevice != null) {
@@ -142,16 +142,10 @@ public class ShooterLimelight extends SubsystemBase {
   }
 
   public Optional<Transform2d> getRobotToScoringLocation() {
-    if (getPipeline() != Constants.limelightPipeline.TAG_PIPELINE) {
-      setPipeline(Constants.limelightPipeline.TAG_PIPELINE);
-    }
     return robotToScoringLocation;
   }
 
   public Optional<Transform2d> checkForTag() {
-    if (getPipeline() != Constants.limelightPipeline.TAG_PIPELINE) {
-      setPipeline(Constants.limelightPipeline.TAG_PIPELINE);
-    }
     if (!LimelightHelpers.getTV(ShooterLimelightConstants.SHOOTER_LIMELIGHT_NAME)) {
       robotToScoringLocation = Optional.empty();
       return Optional.empty();
--- a/src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
+++ b/src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
@@ -1,0 +1,514 @@
+// Code from team 3005
+
+package frc.robot.subsystems.shooterLimelight;
+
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
```diff
@@ -0,0 +1,404 @@
...
+    return table.getEntry("ts").getDouble(0.0);
+  }
+
+  /**
+   * @return latency - The pipeline’s latency contribution in seconds. Add at least 11ms for image
+   *     capture latency.
+   */
+  public double getLatency() {
+    if (m_simDevice != null) {
+      return m_latency.get() + kImageCaptureLatency;
```
balghane, line 280: capture latency is `cl` in the table

balghane: Check in intake too

src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
```diff
@@ -0,0 +1,404 @@
...
+    Rotation2d rot = Rotation2d.fromDegrees(-botPoseTargetSpace.getRotation().getY());
+
+    System.out.println("Tag at " + -botPoseTargetSpace.getRotation().getY() + " deg");
+    return new Transform2d(translation, rot);
+  }
+
+  private static boolean validScoringTag(double tagId) {
+    long tagIdRounded = Math.round(tagId);
+    return (tagIdRounded == 3 || tagIdRounded == 4 || tagIdRounded == 7 || tagIdRounded == 8);
+  }
```
balghane, line 119: In the limelight config (web UI), we should set these as the only valid ones so that the bot pose is only from those tags

src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
```diff
@@ -0,0 +1,404 @@
...
+  private Transform2d getRobotToScoringLocation(Pose3d targetPoseRobotSpace) {
+    Transform2d targetFromBot = getBotFromTarget(targetPoseRobotSpace);
+    return targetFromBot;
+  }
+
+  public Optional<Transform2d> getRobotToScoringLocation() {
+    return robotToScoringLocation;
+  }
+
+  public Optional<Transform2d> checkForTag() {
```
balghane, line 148: I don't think this will work consistently. The "side" speaker tag might be closer and it would be consistently detected over the other one. We should use the megatag pose and use that to point at the speaker

balghane: Or you can iterate through the tag detections and use the center one

tails618: What is the "megatag pose"?

src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
tails618: I probably shouldn't change it for if it's a sim device, right? Just real?

balghane: Ya I suppose

src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
tails618: Done

src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
tails618: Reworked in `basel/integration-v4`

src/main/java/frc/robot/subsystems/shooterLimelight/ShooterLimelight.java
tails618: Pushed

PR #41, opened 2024-02-22
Add rumbleBrieflyCommand when intaking

tails618:
Rumbles the controller when the driver can release the button after intaking

### DIFF
--- a/src/main/java/frc/robot/commands/IntakeSequence.java
+++ b/src/main/java/frc/robot/commands/IntakeSequence.java
@@ -48,9 +48,13 @@ public class IntakeSequence extends SequentialCommandGroup {
         new WaitUntilCommand(elevator::atDesiredPosition),
         new WaitUntilCommand(intake::atDesiredIntakePosition),
         new InstantCommand(intake::startRollers),
-        new ParallelCommandGroup(
+        new ParallelDeadlineGroup(
             Conveyor.receive(conveyor),
-            new WaitUntilCommand(() -> intake.beamBreakSensor.get())), // TODO test true or false
+            new SequentialCommandGroup(
+                new WaitUntilCommand(() -> intake.beamBreakSensor.get())), // TODO test true or false
+                new InstantCommand(intake::stopRollers, intake),
+                new InstantCommand(() -> intake.setDesiredIntakePosition(IntakePosition.STOWED)),
+            )),
         new InstantCommand(intake::stopRollers, intake),
         new InstantCommand(() -> intake.setDesiredIntakePosition(IntakePosition.STOWED)),
         rumbleBrieflyCommand);
--- a/src/main/java/frc/robot/commands/IntakeSequence.java
+++ b/src/main/java/frc/robot/commands/IntakeSequence.java
@@ -1,5 +1,6 @@
 package frc.robot.commands;
 
+import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.ConditionalCommand;
 import edu.wpi.first.wpilibj2.command.InstantCommand;
 import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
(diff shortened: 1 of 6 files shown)


src/main/java/frc/robot/commands/IntakeSequence.java
```diff
@@ -41,12 +48,11 @@ public IntakeSequence(Intake intake, Elevator elevator, Conveyor conveyor, Shoot
...
-            () -> {
-              intake.stopRollers();
-            },
-            intake),
-        new InstantCommand(() -> intake.setDesiredIntakePosition(IntakePosition.STOWED)));
+        new ParallelCommandGroup(
+            Conveyor.receive(conveyor),
+            new WaitUntilCommand(() -> intake.beamBreakSensor.get())), // TODO test true or false
+        new InstantCommand(intake::stopRollers, intake),
+        new InstantCommand(() -> intake.setDesiredIntakePosition(IntakePosition.STOWED)),
```
balghane: I don't think we want to require the beam break to be triggered. If it gets into the conveyor then we should also be done. How about 
```suggestion
        new ParallelDeadlineGroup(
            Conveyor.receive(conveyor),
            new SequentialCommandGroup(
                new WaitUntilCommand(() -> intake.beamBreakSensor.get())), // TODO test true or false
                new InstantCommand(intake::stopRollers, intake),
                new InstantCommand(() -> intake.setDesiredIntakePosition(IntakePosition.STOWED)),
            )),
        new InstantCommand(intake::stopRollers, intake),
        new InstantCommand(() -> intake.setDesiredIntakePosition(IntakePosition.STOWED)),
```

src/main/java/frc/robot/commands/autos/OneFiveLeave.java
```diff
@@ -23,17 +24,19 @@ public OneFiveLeave(
...
     final boolean FIELD_RELATIVE = true;
     final double WAIT_BEFORE_DRIVE_SEC = 1.0;
+    Command rumbleCommand = rumbleBrieflyCommand;
 
     addRequirements(intake, elevator, shooter, conveyor, drive);
     addCommands(
         new ScoreThisNote(intake, elevator, shooter, conveyor),
         new ParallelCommandGroup(
-            new IntakeSequence(intake, elevator, conveyor, shooter)
+            new IntakeSequence(intake, elevator, conveyor, shooter, rumbleCommand)
```
balghane: Maybe make the auto ones a no-op? We don't want to rumble in auto

avakimc: just wondering, what does "no-op" mean?

avakimc (commented):
same comments as basel

src/main/java/frc/robot/commands/autos/OneFiveLeave.java
balghane: no operation. Pass in an empty command so it doesn't rumble in auto

src/main/java/frc/robot/commands/IntakeSequence.java
voidSevenSevenSix: What Basel has written looks correct

tails618: If I'm correct about where it is, the sensor will be triggered before `Conveyor.receive(conveyor)` is done, so I believe the intake will stop and go to stowed position prematurely?

src/main/java/frc/robot/commands/autos/OneFiveLeave.java
tails618: Changed

src/main/java/frc/robot/commands/IntakeSequence.java
balghane: Intake can go in before conveyor is fully done

PR #42, opened 2024-02-23
Smartly Go Home after AmpScore (pulled from main, so may be outdated)

avakimc:


src/main/java/frc/robot/commands/AmpScore.java
```diff
@@ -1,26 +1,48 @@
...
-        // TODO "smartly" go home (e.g. go home once we've driven away from the wall)
-        );
+        Conveyor.scoreTrapOrAmp(conveyor),
+        new DriveDistance(
+            drive,
+            DriveCal.MEDIUM_LINEAR_SPEED_METERS_PER_SEC,
+            NO_X_MOVEMENT,
+            Y_MOVE_18_INCHES_AWAY_FROM_AMP,
+            FIELD_RELATIVE,
+            matchState),
```
balghane, line 45: We shouldn't take over drive, just observe it

balghane (changes requested):

src/main/java/frc/robot/commands/AmpScore.java
avakimc: ohhh, is that why nick mentioned something about the encoders?

balghane: IDK what Nick said but the drive does keep track of the odometry

PR #43, opened 2024-02-23
Attempts to Keep Robot in Same Position After TeleopInit() (pulled from main so may be out of date in some areas)

allison324:


src/main/java/frc/robot/subsystems/elevator/Elevator.java
```diff
@@ -171,6 +173,16 @@ private void controlPosition(ElevatorPosition pos) {
...
+    this.setRealDesired = true;
+  }
+
+  public void setRealDesiredToFalse() {
+    this.setRealDesired = false;
+  }
+
+  public void resetElevatorControllers() {
+    this.noteScoringElevatorController.reset(noteScoringElevatorController.getSetpoint());
+    this.chainGrabberElevatorController.reset(chainGrabberElevatorController.getSetpoint());
```
balghane, line 185: Controllers should only ever be reset to the current measurement

src/main/java/frc/robot/subsystems/elevator/Elevator.java
```diff
@@ -199,7 +211,9 @@ public boolean elevatorAboveInterferenceZone() {
 
   @Override
   public void periodic() {
-    controlPosition(desiredPosition);
+    if (setRealDesired) {
+      controlPosition(desiredPosition);
+    }
   }
```
balghane, line 217: If the elevator is above the intake, then robot gets disabled, then elevator drifts down onto the intake. I'm a little worried about enabling and clicking home, which would deploy the intake. Intake would be in contact with the elevator while it moves which could be bad. We should take a look at that.

PR #48, opened 2024-03-09
Changes from AIM 2/26-2/27 and Pittsburgh

tails618:
* Replicate existing code from `ava/feb-26-aim` (note that this was done manually so I would NOT expect it to merge easily if that is necessary for whatever reason)
* Remove most commented out code
* Add auton chooser to SmartDashboard
* Updated/cleaned up match state logic by putting it in a util
* Cleaned up booleans being used for shooter logic
* Cleaned up a few prints (none removed, but there shouldn't be any printing out any raw values with no context)

I tried to keep commits pretty clear, so you can review in chunks. I would recommend reviewing the changes from `ava/feb-26-aim` (plus associated format commit) separately from the actual clean-up, to make it easier on yourself.


src/main/java/frc/robot/Robot.java
```diff
@@ -177,19 +180,18 @@ public void simulationInit() {}
   @Override
   public void simulationPeriodic() {}
 
+  /**
+   * Sets whether it is a real match by checking if the match time is more than 1s, and whether we
+   * are blue from the driver station data (defaults to true).
+   */
   private void setRobotContainerMatchState() {
```
balghane: rename since it's doing it in here

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -131,12 +140,38 @@ public RobotContainer() {
...
+    autonChooser.setDefaultOption(
+        "Score two",
+        new ScoreTwoNotes(
+            intake, elevator, shooter, conveyor, drive, matchState, shooterLimelight));
+    autonChooser.addOption(
+        "Score four from center",
+        new ScoreFourFromCenterLine(drive, intake, elevator, shooter, conveyor, shooterLimelight));
+    autonChooser.addOption(
+        "Score two with center note",
+        new TwoWithCenterNote(drive, intake, elevator, shooter, conveyor, shooterLimelight));
```
balghane: Remove since we haven't ever tried it

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -150,75 +185,143 @@ public RobotContainer() {
...
         .rightTrigger()
+        .onFalse(new GoHomeSequence(intake, elevator, shooter, conveyor, false, false));
+    driverController
+        .leftTrigger()
         .onTrue(
             new ConditionalCommand(
-                new SpeakerShootSequence(conveyor, shooter).finallyDo(conveyor::stopRollers),
+                new ConditionalCommand(
+                    new SpeakerShootSequence(
+                            conveyor, shooter, elevator, drive, shooterWaitUntilRotated)
```
balghane: We should be passing in a boolean supplier, this sends in the initial value

balghane (approved):

PR #55, opened 2024-06-22
demo mode additions (commented out though)

avakimc:
- RobotContainer: noteTooCloseThreshold is changed to 0.5 for auto intake
- RobotContainer: drive.throttleSpeed() is set to true
- RobotContainer: ClimbPrep button binding is commented out
- AmpPrepScoreSequence: drive train doesn't automatically rotate
- SpeakerPrepScoreSequence: drive train doesn't automatically rotate
- FeedPrepScore: drive train doesn't automatically rotate


jacobfeder (approved):
sid, jacob, and ava all hope this works

PR #61, opened 2024-09-22
Added vision to auto, and added an auto to test

voidSevenSevenSix:


src/main/java/frc/robot/commands/AutoVisionCommand.java
```diff
@@ -0,0 +1,78 @@
...
+  public Optional<Pair<Pose2d, Pose2d>> poses = Optional.empty();
+  public AutoVisionCommand(
+      DriveSubsystem drive,
+      Intake intake,
+      Conveyor conveyor,
+      Elevator elevator,
+      Shooter shooter,
+      Lights lights,
+      IntakeLimelight intakeLimelight) {
+    addRequirements(drive, intake, conveyor, elevator);
```
avakimc, line 39: shouldn't the shooter be included as a requirement?

src/main/java/frc/robot/commands/AutoVisionSequence.java
```diff
@@ -0,0 +1,46 @@
...
+public class AutoVisionSequence extends SequentialCommandGroup {
+  private PathPlannerPath path;
+  public AutoVisionSequence(DriveSubsystem drive,
+      Intake intake,
+      Conveyor conveyor,
+      Elevator elevator,
+      Shooter shooter,
+      Lights lights,
+      IntakeLimelight intakeLimelight) {
+        addRequirements(drive, intake, conveyor, elevator);
```
avakimc, line 29: again, don't you need to add the shooter as a requirement?

src/main/java/frc/robot/commands/AutoVisionCommand.java
```diff
@@ -0,0 +1,78 @@
...
+                        robotPoseAtDetection.plus(
+                            new Transform2d(
+                                poseAtDetectionToNote,
+                                Rotation2d.fromDegrees(latestNoteDetection.yawAngleDeg)));
+                    List<Translation2d> bezierPoints = PathPlannerPath.bezierFromPoses(
+                      curPose,
+                      goalPose
+                    );
+                    poses = Optional.of(new Pair<Pose2d, Pose2d>(curPose, goalPose));
+                    path = Optional.of(new PathPlannerPath(bezierPoints, new PathConstraints(5.0, 6.0, 2*Math.PI, 4*Math.PI), new GoalEndState(2.5, Rotation2d.fromDegrees(latestNoteDetection.yawAngleDeg))));
```
avakimc, line 72: can you make these variables in constants or cal

src/main/java/frc/robot/commands/AutoVisionCommand.java
```diff
@@ -0,0 +1,78 @@
...
+                  }
+                }), 
+                path.isPresent()?drive.followTrajectoryCommand(path.get(), false):new InstantCommand()
```
avakimc: maybe put a print in the instant command that a path couldn't be made

src/main/java/frc/robot/commands/AutoVisionSequence.java
```diff
@@ -0,0 +1,46 @@
...
+        AutoVisionCommand autoVisionCommand = new AutoVisionCommand(drive, intake, conveyor, elevator, shooter, lights, intakeLimelight);
+        addCommands(
+          autoVisionCommand, 
+          autoVisionCommand.poses.isEmpty()?new InstantCommand():new SequentialCommandGroup(
+            new InstantCommand(()->{
+              List<Translation2d> bezierPoints = PathPlannerPath.bezierFromPoses(
+                autoVisionCommand.poses.get().getSecond(),
+                autoVisionCommand.poses.get().getFirst()
+              );
+              path = new PathPlannerPath(bezierPoints, new PathConstraints(5.0, 6.0, 2*Math.PI, 4*Math.PI), new GoalEndState(1, autoVisionCommand.poses.get().getSecond().getRotation()));
```
avakimc: shouldn't goal end state use poses.get().getFirst() because the first pose is the curPose which is the one we want to drive back to?

avakimc (commented):
some minimal comments, may be incorrect on some

src/main/java/frc/robot/commands/AutoVisionCommand.java
voidSevenSevenSix: i don't think so

src/main/java/frc/robot/commands/AutoVisionSequence.java
voidSevenSevenSix: i don't think so

src/main/java/frc/robot/commands/AutoVisionCommand.java
voidSevenSevenSix: done

src/main/java/frc/robot/commands/AutoVisionCommand.java
voidSevenSevenSix: later

src/main/java/frc/robot/commands/AutoVisionSequence.java
voidSevenSevenSix: fixed

src/main/java/frc/robot/commands/AutoVisionCommand.java
tails618: Pretty sure it should, because this command actually uses the shooter as far as I can tell. I don't remember how requirement inheritance works (i.e. `IntakeSequence` requires the shooter, so I *think* this will automatically require it?) but it's probably best practice to explicitly state it at the top – if for no reason other than clarity for the reader. 

~~it would be really embarrassing for me if i was wrong~~

